PR #38, opened 2026-06-27
Add standalone macro editor page

IgorArkhipov:
### Summary
- Adds a standalone Macros page to the Settings dropdown when the connected keyboard exposes macro slots.
- Reuses the existing macro editor so macros can be selected, viewed, and edited without assigning the macro to a key.
- Splits macro editor commit behavior so the key picker still assigns `M<n>`, while the standalone page saves without changing the currently selected keycode.

### Verification
- `git diff --check`
- `/Users/igor.arkhipov/.cargo/bin/cargo build --quiet`
- `/Users/igor.arkhipov/.cargo/bin/cargo test --quiet` passes: 46 tests.

### Кратко
- Добавляет отдельную страницу Macros в меню настроек, если подключенная клавиатура поддерживает макросы.
- Переиспользует существующий редактор макросов, чтобы макросы можно было выбирать, просматривать и редактировать без назначения макроса на клавишу.
- Разделяет поведение сохранения: пикер клавиш по-прежнему назначает `M<n>`, а отдельная страница сохраняет макрос без изменения выбранного keycode.

### Проверка
- `git diff --check`
- `/Users/igor.arkhipov/.cargo/bin/cargo build --quiet`
- `/Users/igor.arkhipov/.cargo/bin/cargo test --quiet` проходит: 46 тестов.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -115,6 +115,8 @@ mod layout_view;
 mod live_features_settings_ui;
 #[path = "ui/magic_settings.rs"]
 mod magic_settings_ui;
+#[path = "ui/macro_settings.rs"]
+mod macro_settings_ui;
 #[path = "ui/matrix_tester.rs"]
 mod matrix_tester;
 #[path = "ui/module_settings.rs"]
--- a/src/app_state.rs
+++ b/src/app_state.rs
@@ -1380,6 +1380,7 @@ pub(crate) enum SettingsTab {
     AppSettings,
     MatrixTester,
     UniversalSymbolsSetup,
+    Macros,
     TextExpander,
     AutoShift,
     Rgb,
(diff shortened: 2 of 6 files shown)


IgorArkhipov:
Rebased this branch onto current `origin/main`.

The conflict in `src/keycode_picker_macro.rs` was only in the test module, so I kept the lossless macro bytecode round-trip coverage from #32 together with this PR’s standalone macro editor commit-mode tests.

The PR is now clean against main and CI is green (4/4).

kissetfall:
Thanks. Keeping this one open, but not merging it for now. The standalone macro editor is still a product question for us: we may add it later, but we would prefer to avoid a separate page if we can keep the macro workflow simple enough inside the existing picker/settings structure.

kissetfall:
Thank you for putting this together and for exploring a standalone macro editing workflow.

After extensive review and discussion, we have decided not to implement a separate macro editor page. Although the idea is reasonable, we do not believe the additional Settings surface and maintenance cost provide enough product value compared with the existing macro editing flow.

This is a product-scope decision rather than a request for further revisions, so we are closing the PR without merging. No additional work is needed on this branch. Thank you again for the contribution.

Issue #59, opened 2026-06-29
High CPU usage when minimized to the system tray


psycha0s:
After minimizing Entropy to the system tray, the process starts using 100% of a single CPU core.

**Steps to reproduce:**
- Launch Entropy on Windows 11
- Enable the option "Close to tray" in "App Settings"
- Minimize the application to the system tray
- Observe CPU usage via Task Manager

**Expected behavior:** CPU usage should remain idle when the app is in the background

**Actual behavior:** CPU usage remains unusually high when the app is minimized to the tray

<img width="875" height="334" alt="Image" src="https://github.com/user-attachments/assets/6a7a64db-e89b-4af3-a259-4c6d2e220099" />
<img width="1207" height="1233" alt="Image" src="https://github.com/user-attachments/assets/92d19f37-b1ed-4f82-9dc3-e588bdd76906" />

IgorArkhipov:
This should be addressed by #54.

That PR throttles background updates while the app is minimized/hidden to tray, which matches this report about high CPU usage after closing the main window with tray mode enabled.

psycha0s:
Just tested the version 0.2.0 and the issue is still here.

kissetfall:
Reopening: this still reproduces on Windows 11 in v0.2.0 according to the latest test report. #54 reduced background update frequency but did not fully address the tray/minimized CPU spin, so this needs another pass.

IgorArkhipov:
> Reopening: this still reproduces on Windows 11 in v0.2.0 according to the latest test report. [#54](https://github.com/ergohaven/entropy/pull/54) reduced background update frequency but did not fully address the tray/minimized CPU spin, so this needs another pass.

Thanks for retesting!
I’ll restart the investigation from the actual Windows 11 tray path and separate repaint/event-loop wakeups from device polling, text-expander maintenance, and tray restore polling.

If convenient, could you confirm whether the behavior changes with:
1. no keyboard connected versus USB/Bluetooth;
2. restoring the main window;
3. using the minimize button versus closing the window to tray?

Those distinctions should narrow the remaining scope considerably.

psycha0s:
Okay, so I tried [this artifact](https://github.com/ergohaven/entropy/actions/runs/29162249098).
The issue occurs in the following scenarios:
- **USB connection:** The app is closed to the system tray - issue exists.
(Works correctly when the app is in the foreground or minimized.)
[...]

PR #64, opened 2026-07-12
Fix high CPU load on Windows 11

psycha0s:
I tried to investigate #59 on my system and was able to tame it. Since I'm not familiar with Rust and egui, I had to use AI, so take everything with a big grain of salt. It looks like an ugly hack to me, but it does resolve the high CPU load issue in both Bluetooth and USB connection modes, whether the app is in the foreground or hidden in the system tray. I don't expect this to be merged as-is - I mainly needed a working fix for myself, but I hope it can give you some ideas how to implement a proper fix.

### DIFF
--- a/src/app_init.rs
+++ b/src/app_init.rs
@@ -63,6 +63,7 @@ impl EntropyApp {
             import_progress_title: String::new(),
             import_progress_body: String::new(),
             dark_mode,
+            last_applied_dark_mode: None,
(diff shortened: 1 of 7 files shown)


kissetfall:
Closing as superseded by #69, which uses the eframe lifecycle fix and bounded idle repaint scheduling and has completed Windows runtime validation.

PR #77, opened 2026-07-13
Persist imported layer names to firmware

ImmortalDragonm:
### Problem

After importing an `.entlayout`, layer names appear correctly, but once the app is closed and reopened only the first layer name survives — the rest fall back to firmware defaults / numbers.

### Cause

Two issues combine:

1. `apply_entlayout_firmware_state` writes the keymap, encoders, macros, combos, tap dance, key overrides and alt-repeat back to the keyboard, but **never writes layer names** (qsid `200 + layer`). Imported names only reach the local cache file (`save_layer_names`).
2. On the next connect, `device_connect_apply` reads names from firmware and used an **all-or-nothing** rule (`has_firmware_layer_names`): if the firmware reported *any* non-default name, it treated firmware as authoritative for *every* layer and ignored the locally-saved names entirely. Because the firmware kept a real name for layer 0 (e.g. `BASE`), every other layer's imported name was dropped.

### Fix

- Import now writes each imported layer name to `qsid 200 + layer` for Vial devices, so names persist on the keyboard and survive reconnects. Unsupported qsids fail softly and are reported like any other firmware-section failure.
- On connect, locally-saved names are overlaid **per layer** wherever the firmware only reports a default placeholder, instead of all-or-nothing. A real firmware name still wins; other layers keep their saved names. The now-unused `has_firmware_layer_names` helper is removed.

### Testing

- `cargo build`, `cargo test`, `python3 scripts/check_i18n.py` all pass.
- Verified on an Ergohaven Phenom (Vial): the device reports layer-name settings at qsid 200–215; after the fix, imported names round-trip through a reconnect instead of collapsing to the first layer.

https://claude.ai/code/session_0112ag6JdE9PkygMpWweftUA

### DIFF
--- a/src/entlayout.rs
+++ b/src/entlayout.rs
@@ -644,9 +644,39 @@ impl EntropyApp {
         bundle: &EntLayoutFile,
     ) -> Result<(Vec<String>, EntLayoutImportMapping)> {
         let mapping = self.entlayout_import_mapping(bundle)?;
-        let firmware_failures = self.apply_entlayout_firmware_state(bundle, &mapping)?;
+        let mut firmware_failures = self.apply_entlayout_firmware_state(bundle, &mapping)?;
         self.apply_entlayout_local_state(bundle, &mapping)?;
         self.refresh_layer_picker_content_flags();
+
+        // Persist the imported layer names into firmware now that self.layer_names
+        // holds the final target values. The shared writer continues past a
+        // failing layer and reports which ones failed, so stale firmware slots
+        // aren't left to win over the local names on the next reconnect.
+        #[cfg(not(target_arch = "wasm32"))]
+        {
+            let names = self.layer_names.clone();
+            let failed = self.write_layer_names_to_firmware(&names);
+            if !failed.is_empty() {
+                let lang = self.app_settings.language;
+                let layers = failed
+                    .iter()
+                    .map(|layer| layer.to_string())
+                    .collect::<Vec<_>>()
+                    .join(", ");
+                firmware_failures.push(crate::i18n::tr_catalog_format(
+                    lang,
+                    "entlayout.firmware_failure",
+                    &[
+                        (
+                            "section",
+                            crate::i18n::tr_catalog(lang, "entlayout.layer_names"),
+                        ),
+                        ("error", &format!("layers {layers}")),
+                    ],
+                ));
+            }
+        }
+
         Ok((firmware_failures, mapping))
     }
 
@@ -1008,35 +1038,9 @@ impl EntropyApp {
             ));
         }
 
-        // Persist layer names into firmware (qsid 200 + layer). Without this the
-        // imported names live only in the local cache file and get overwritten by
-        // the firmware's own names on the next reconnect. Unsupported qsids simply
-        // return an error, which we treat as "firmware has no layer-name storage".
-        if self.firmware == FirmwareProtocol::Vial {
-            if let Err(err) = (|| -> Result<()> {
-                for (source_layer_idx, name) in bundle.data.layer_names.iter().enumerate() {
-                    let Some(target_layer_idx) =
-                        map_layer_index(source_layer_idx, self.layer_count)
-                    else {
-                        continue;
-                    };
-                    hid.set_qmk_setting_string(200 + target_layer_idx as u16, name)?;
-                }
-                Ok(())
-            })() {
-                failures.push(crate::i18n::tr_catalog_format(
-                    lang,
-                    "entlayout.firmware_failure",
-                    &[
-                        (
-                            "section",
-                            crate::i18n::tr_catalog(lang, "entlayout.layer_names"),
-                        ),
-                        ("error", &err.to_string()),
-                    ],
-                ));
-            }
-        }
+        // Layer names are written after local state sets self.layer_names, via
+        // the shared write_layer_names_to_firmware writer (see apply_entlayout),
+        // so a single failing layer no longer aborts the rest.
 
         if let Err(err) = (|| -> Result<()> {
             for (source_layer_idx, layer_codes) in bundle.data.encoder_keymap.iter().enumerate() {
--- a/src/entlayout.rs
+++ b/src/entlayout.rs
@@ -1008,6 +1008,36 @@ impl EntropyApp {
             ));
         }
 
+        // Persist layer names into firmware (qsid 200 + layer). Without this the
+        // imported names live only in the local cache file and get overwritten by
+        // the firmware's own names on the next reconnect. Unsupported qsids simply
+        // return an error, which we treat as "firmware has no layer-name storage".
+        if self.firmware == FirmwareProtocol::Vial {
+            if let Err(err) = (|| -> Result<()> {
+                for (source_layer_idx, name) in bundle.data.layer_names.iter().enumerate() {
+                    let Some(target_layer_idx) =
+                        map_layer_index(source_layer_idx, self.layer_count)
+                    else {
+                        continue;
+                    };
+                    hid.set_qmk_setting_string(200 + target_layer_idx as u16, name)?;
+                }
+                Ok(())
+            })() {
+                failures.push(crate::i18n::tr_catalog_format(
+                    lang,
+                    "entlayout.firmware_failure",
+                    &[
+                        (
+                            "section",
+                            crate::i18n::tr_catalog(lang, "entlayout.layer_names"),
+                        ),
+                        ("error", &err.to_string()),
+                    ],
+                ));
+            }
+        }
+
         if let Err(err) = (|| -> Result<()> {
             for (source_layer_idx, layer_codes) in bundle.data.encoder_keymap.iter().enumerate() {
                 let Some(target_layer_idx) = map_layer_index(source_layer_idx, self.layer_count)
--- a/src/hid_settings.rs
+++ b/src/hid_settings.rs
@@ -4,6 +4,31 @@ use super::hid_parse::{
 };
 use super::hid_protocol::*;
 use super::HidDevice;
+
+/// Escape `%` as `%%` and pack `value` into at most `budget` bytes, stopping at
+/// a whole-character boundary. Never splits a UTF-8 codepoint or a `%%` pair, so
+/// the firmware never receives a corrupted string.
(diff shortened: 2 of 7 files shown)


src/entlayout.rs
```diff
@@ -1008,6 +1008,36 @@ impl EntropyApp {
...
+        // return an error, which we treat as "firmware has no layer-name storage".
+        if self.firmware == FirmwareProtocol::Vial {
+            if let Err(err) = (|| -> Result<()> {
+                for (source_layer_idx, name) in bundle.data.layer_names.iter().enumerate() {
+                    let Some(target_layer_idx) =
+                        map_layer_index(source_layer_idx, self.layer_count)
+                    else {
+                        continue;
+                    };
+                    hid.set_qmk_setting_string(200 + target_layer_idx as u16, name)?;
```
IgorArkhipov: Nice improvement to firmware persistence!

One edge case: set_qmk_setting_string accepts 27 payload bytes, so a valid multibyte name such as seven emoji can be truncated mid-codepoint.
After reconnect, that corrupted firmware value is considered authoritative over intact local value.

Could we reject oversized escaped payloads or truncate only at valid UTF-8 boundary?
A round-trip test plus unsupported-QSID failure test would protect this path.

src/ui/device_connect_apply.rs
```diff
@@ -202,23 +195,22 @@ impl EntropyApp {
...
-                            .enumerate()
-                            .take(r.layer_count)
-                        {
-                            if !name.trim().is_empty() {
-                                layer_names[idx] = name;
-                            }
+                if let Some(local_layer_names) = load_saved_layer_names(&device_name) {
+                    for (idx, name) in local_layer_names.into_iter().enumerate().take(r.layer_count)
+                    {
+                        if !name.trim().is_empty() && is_default_layer_name(idx, &layer_names[idx]) {
```
IgorArkhipov: Per-layer fallback looks good overall.

One precedence case seems ambiguous: successful firmware read of Main is classified as default text, so stale locally saved name can overwrite it.

Could connect result retain whether each name came from successful QSID read, then apply local fallback only to descriptor placeholders?

Table-driven tests for mixed firmware/local names and default variants would help lock behavior down.

ImmortalDragonm:
Thanks for the review! Addressed both points:

- **Truncation:** `set_qmk_setting_string` now packs the escaped name stopping only at whole characters, so it never splits a multibyte codepoint (emoji/Cyrillic) or a `%%` escape pair. Added unit tests across ASCII / percent / emoji / Cyrillic.
- **Precedence:** the connect result now records, per layer, whether the name came from a successful QSID read. A stored firmware name is authoritative even when it looks like a default (e.g. `Main`); locally saved names only fill layers the firmware left as generated placeholders. Extracted the merge into `resolve_layer_names` with a table-driven test matrix (firmware-wins, placeholder-fill, mixed, extend/truncate).

kissetfall:
Thanks, the two earlier correctness issues look addressed now:

- `set_qmk_setting_string` now truncates at whole UTF-8 characters and keeps escaped `%%` pairs intact.
- Layer-name precedence is now based on per-layer QSID provenance, so a real firmware name such as `Main` is authoritative and stale local names only fill descriptor/default placeholders.

I would not block this on the implementation logic anymore, but the branch is stale against current `main`:

- GitHub reports `mergeStateStatus=DIRTY`.
- Local merge conflicts in `src/ui/device_connect_apply.rs`; the resolution should keep both the existing HID-open-failure test from `main` and the new `resolve_layer_names` tests from this PR.
- After resolving that, the branch still defines a second `mod tests` in `src/hid_settings.rs`; the new truncation tests need to move into the existing test module.

With those mechanical rebase fixes applied in a temporary worktree, `cargo test --locked` passes locally (199/199), `python3 scripts/check_i18n.py` passes, and `git diff --check` passes. Please rebase onto current `main` and rerun CI; after that this should be ready for approval/merge from my side.

ImmortalDragonm:
Rebased onto current `main`. Resolved the `device_connect_apply.rs` conflict keeping both the existing HID-open-failure test and the new `resolve_layer_names` tests, and merged the truncation tests into the existing `mod tests` in `hid_settings.rs` (no more second module). `cargo test` 199/199, `check_i18n.py` and `cargo fmt --check` on the changed files all pass.

src/hid_settings.rs
```diff
@@ -238,12 +263,14 @@ impl HidDevice {
...
-        let safe_value = value.replace('%', "%%");
-        let bytes = safe_value.as_bytes();
+        // Escape '%' as '%%' and pack into the fixed payload, truncating only at
+        // whole characters — never mid-codepoint and never splitting a '%%' pair,
+        // either of which would leave a corrupted string in firmware.
         let max_len = cmd.len().saturating_sub(4);
-        let copy_len = bytes.len().min(max_len.saturating_sub(1));
-        cmd[4..4 + copy_len].copy_from_slice(&bytes[..copy_len]);
-        cmd[4 + copy_len] = 0;
+        let budget = max_len.saturating_sub(1); // reserve one byte for the null terminator
```
IgorArkhipov: Thanks for making the 27-byte HID payload UTF-8-safe.

One smaller boundary remains: Ergohaven firmware stores layer names in a 16-byte buffer, leaving 15 data bytes. Longer emoji or Cyrillic names can pass this helper but still be split by firmware; reconnect then treats replacement text as authoritative.

Could we enforce both limits - 27 escaped wire bytes and 15 decoded label bytes - or reject oversized names?

src/hid_settings.rs
```diff
@@ -381,6 +408,39 @@ impl HidDevice {
 mod tests {
     use super::*;
 
+    #[test]
+    fn truncate_short_ascii_is_unchanged() {
+        assert_eq!(truncate_qmk_string_payload("BASE", 27), b"BASE");
```
IgorArkhipov: Nice helper-level boundary tests!

One part of the earlier request remains open: these tests do not call production set_qmk_setting_string / get_qmk_setting_string, so command framing, unsupported-QSID responses, firmware-shaped readback, and import continuation can regress while this suite stays green.

Could we add a scripted HID seam, or extract production-used packet helpers, then cover round-trip plus unsupported QSID and a middle-layer failure?

src/entlayout.rs
IgorArkhipov: This `?` exits the entire layer-name loop after the first failed SET, so later QSIDs are never attempted. 

Local import still saves every requested name, but reconnect treats stale names in those untouched firmware slots as authoritative.

Current main also has sync_layer_names_to_firmware; could we rebase and share one per-layer writer that continues independent layers and aggregates failures?

ImmortalDragonm:
Addressed the two correctness points and rebased onto current `main`:

- **15-byte firmware buffer:** `truncate_qmk_string_payload` now honours *both* limits — the 15 decoded bytes of the firmware label buffer and the 27 escaped wire bytes — still stopping only at whole characters. Added decoded-limit tests (emoji/ASCII) plus one using the exact production 15/27 bounds.
- **`?` aborting the loop / shared writer:** extracted `write_layer_names_to_firmware` from `main`'s `sync_layer_names_to_firmware`; it writes per layer, continues past a failing layer, skips names already matching, and returns the failed layer indices. The import removes its bail-on-first loop and calls this shared writer after local state, aggregating any failures into the import report.

Still open: the request to exercise production `set/get_qmk_setting_string` through a scripted HID seam. That's a larger refactor (introducing a mockable transport); happy to do it as a focused follow-up so this correctness fix isn't held on the test-harness work — let me know your preference.

kissetfall (changes requested):
Thanks for the work here. Requesting changes before this can merge.

1. Rebase onto the current `main` and resolve the conflict in `src/ui/device_connect_apply.rs`.
2. Fix the false-success path in `write_layer_names_to_firmware`. At the moment any `get_qmk_setting_string` error is treated as “storage unsupported”: layer 0 stops probing, while later-layer read errors are only logged. A transient USB/Bluetooth read failure therefore skips the `SET`, returns no failed layer, and lets `.entlayout` import report success even though the firmware still contains the old name. Unsupported storage must be distinguished from transport/read failures, or the affected layer must be returned as failed. Success should only be reported when every non-empty name either already matched or was written successfully.
3. Add focused tests around the production HID path, not only the pure name/truncation helpers. Please cover unsupported layer-name storage, a transient first-layer read failure, a middle-layer read failure, a `SET` failure followed by successful later layers, and persistence after reconnect.

Ready for re-review when the branch merges cleanly with current `main`, the failure reporting is accurate, and the new regression tests pass.

ImmortalDragonm:
Addressed the round-3 review:

**False-success fix.** A read/write error on a *supported* qsid was being treated as "firmware has no layer-name storage" — layer 0 aborted the whole sync and later layers were only logged, so a transient USB/BT hiccup let an import report success while the firmware kept the old names. The sync now gates on the setting ids the connect probe advertised (`supported_qmk_settings`, plumbed through `ConnectResult` → app state): an advertised-but-failing id is reported as a failed layer, and only a genuinely absent id is skipped. Success now means every non-empty name either already matched or was written back — never a masked transport error.

**Testable production path.** Extracted the sync loop behind a `LayerNameStore` seam (`sync_layer_names_to_store`), with the real HID device as one impl and an in-memory mock as the other. New tests cover: unsupported storage (no wire traffic, not a failure), transient first-layer read failure (reported, later layers still written — the exact regression), middle-layer read failure, SET failure with later layers still persisted, matching/empty names skipped, and idempotent re-sync after reconnect.

Rebased onto current `main` (reusing its `supported_qmk_settings` probe and layer-name precedence). `cargo test` 267 passing, `cargo fmt --check` and `check_i18n.py` clean.

kissetfall (approved):
Re-reviewed head f0d0631. The false-success path is fixed: advertised layer-name QSIDs are carried into the live connection, unsupported storage is skipped without wire traffic, and a read/write error on a supported QSID is reported for that layer while later layers continue.

The extracted LayerNameStore path now covers unsupported storage, first/middle read failures, SET failure with continued writes, idempotent reconnect, and matching/empty names. Local verification passed 267/267 tests, scoped rustfmt, i18n, diff-check, and a release build. GitHub has no hosted checks configured for this branch, but the PR is mergeable and the requested regression coverage passes locally.

Approved.

PR #78, opened 2026-07-13
Add PDF export for keyboard layers

ImmortalDragonm:
### Summary

Adds **PDF** as a third format on the "Export layout image" page, alongside PNG and SVG, so a layout can be printed directly. Each selected layer goes on its own A4 page.

### Details

- New `LayoutImageExportFormat::Pdf` variant; the format dropdown, save dialog (`.pdf`, "PDF document") and export button all handle it. When PDF is selected the page shows a hint: "Printable PDF: each selected layer on its own A4 page".
- `render_layout_pdf` reuses the existing raster render (`render_layout_image`) per layer, then embeds each page image via a small self-contained PDF writer:
  - PDF 1.4, one flate-compressed `DeviceRGB` image XObject per page (compression via `flate2`, already a dependency — no new crates).
  - Auto orientation: landscape A4 for wide boards, portrait otherwise; image centered and scaled to fit the margins, never upscaled past 1 pt/pixel.
  - No font embedding needed — layer titles are already rasterized into the page image, so Cyrillic layer names render correctly.

### Testing

- Unit tests added: page-count/orientation, empty-selection rejection, xref-offset integrity. `cargo test`, `python3 scripts/check_i18n.py` pass.
- Manually validated a generated 3-page PDF with `pdfinfo` (3 A4 pages) and Ghostscript (renders all pages); inspected the rasterized pages for correct orientation, centering and colors.

### Notes

- All new firmware/heavy code is `cfg(not(target_arch = "wasm32"))`, matching the existing PNG/SVG export gating; the wasm build keeps the button disabled.

https://claude.ai/code/session_0112ag6JdE9PkygMpWweftUA


src/app_state.rs
```diff
@@ -2605,6 +1419,7 @@ pub(crate) enum LayoutImageExportFormat {
     #[default]
     Png,
     Svg,
+    Pdf,
```
IgorArkhipov, line 2608: Could we keep this persisted format backward-compatible?

Selecting PDF writes "pdf" into AppSettings;
v0.2.0 cannot deserialize it, falls back to all defaults, then writes those defaults back on exit.

It also enters `.entsettingsf` while bundle remains version 1.

Could we use an optional sidecar field ignored by older builds, or introduce a properly versioned settings shape, with downgrade/import fixtures?

src/ui/layout_image_export.rs
```diff
@@ -981,6 +1005,21 @@ impl EntropyApp {
...
+    /// raster render, so the file is ready for printing as-is.
+    #[cfg(not(target_arch = "wasm32"))]
+    fn render_layout_pdf(
+        &self,
+        layout: &KeyboardLayout,
+        selected_layers: &[usize],
+    ) -> anyhow::Result<Vec<u8>> {
+        let mut pages = Vec::with_capacity(selected_layers.len());
+        for layer_idx in selected_layers.iter().copied() {
+            pages.push(self.render_layout_image(layout, &[layer_idx])?);
```
IgorArkhipov: Could we avoid retaining every rendered page here?

A valid large 32-layer export can keep roughly 18.3 GB of RGBA pages alive before RGB and compression buffers.

An aggregate pixel budget plus rendering, encoding, and dropping one page at a time would keep this failure recoverable.
An oversized 32-layer test would help too.

src/ui/layout_image_export.rs
```diff
@@ -1977,3 +2016,173 @@ fn blend_pixel(image: &mut RgbaImage, x: i32, y: i32, color: Rgba<u8>, coverage:
...
+#[cfg(not(target_arch = "wasm32"))]
+const PDF_PAGE_MARGIN_PT: f32 = 28.35;
+
+/// Minimal PDF writer: one A4 page per image, auto orientation, image
+/// centered and scaled to fit inside the page margins (never upscaled
+/// past 1 pt per pixel). Images embed as flate-compressed DeviceRGB
+/// XObjects, so no font embedding is needed for Cyrillic layer names —
+/// all text is already rasterized into the page image.
+#[cfg(not(target_arch = "wasm32"))]
+fn build_pdf_from_images(images: &[RgbaImage]) -> anyhow::Result<Vec<u8>> {
```
IgorArkhipov: This writer is nicely self-contained.

Would you be open to moving the A4 constants, build_pdf_from_images, and its tests into a focused PDF module?

This adds 209 lines to an already large UI module, and extraction would make the streaming change above easier to implement and review.

src/ui/layout_image_export.rs
```diff
@@ -1977,3 +2016,173 @@ fn blend_pixel(image: &mut RgbaImage, x: i32, y: i32, color: Rgba<u8>, coverage:
...
+        let page_id = 3 + page * 3;
+        let contents_id = page_id + 1;
+        let image_id = page_id + 2;
+
+        let (page_w, page_h) = if image.width() >= image.height() {
+            (PDF_A4_LONG_PT, PDF_A4_SHORT_PT)
+        } else {
+            (PDF_A4_SHORT_PT, PDF_A4_LONG_PT)
+        };
+        let scale = ((page_w - PDF_PAGE_MARGIN_PT * 2.0) / image.width() as f32)
```
IgorArkhipov: Could we add assertions for the generated content matrix?

Current tests cover page count and orientation, but scaling, margins, centering, and the 1 pt/pixel no-upscale rule could regress while staying green.

Small-image and oversized-image cases should cover both branches.

src/ui/layout_image_export.rs
```diff
@@ -1977,3 +2016,173 @@ fn blend_pixel(image: &mut RgbaImage, x: i32, y: i32, color: Rgba<u8>, coverage:
...
+    fn pdf_export_rejects_empty_layer_selection() {
+        assert!(build_pdf_from_images(&[]).is_err());
+    }
+
+    #[test]
+    fn pdf_xref_offsets_point_at_objects() {
+        let image = RgbaImage::from_pixel(4, 4, Rgba([0, 128, 255, 255]));
+        let pdf = build_pdf_from_images(&[image]).expect("pdf should build");
+        let text = String::from_utf8_lossy(&pdf);
+        let xref_pos = text.rfind("\nxref\n").expect("xref table present") + 1;
```
IgorArkhipov: Could we strengthen this test to follow the trailer pointer instead of locating xref directly?

As written, the loop can inspect zero or partial rows and still pass.

Asserting that startxref matches the raw xref offset, row count matches the declared object count, and every non-free offset is valid would make the integrity check stronger.

ImmortalDragonm:
Thanks for the thorough review! All five points addressed:

- **Backward-compatible settings:** `LayoutImageExportState` now persists through a mirror shape whose `format` only ever holds `png`/`svg`; PDF is recorded in an extra `export_pdf` flag that pre-PDF builds silently ignore. So v0.2.0 no longer fails to deserialize and reset every setting. Added round-trip tests for all formats, a legacy-json (no sidecar) read test, and assertions that a PDF selection serializes as `format:"png"` + `export_pdf:true`. This also covers the value travelling inside `.entsettings`.
- **Memory:** the export now streams — `build_layer_pdf` pulls one page image at a time via a callback, encodes it and drops it before the next, so peak memory is a single page rather than all 32.
- **Module extraction:** the writer, A4 constants and tests moved into a dedicated `src/pdf.rs`.
- **Content-matrix tests:** page placement is now a pure function with assertions for the no-upscale rule, margin-fit, aspect preservation and centering, on both small and oversized images.
- **xref test:** now follows the trailer's `startxref` pointer, checks the declared object count, and validates every non-free entry, so it can't pass on a partial table.

Rebased onto latest main.

kissetfall:
Thanks, the implementation logic looks good to me now. The earlier review points are addressed:

- PDF selection persists backward-compatibly as `format:"png"` plus `export_pdf:true`, so old builds should not reset all settings.
- PDF generation streams one rendered layer at a time instead of retaining all page images.
- The PDF writer/tests moved into `src/pdf.rs`.
- Placement tests now cover no-upscale, margins, aspect ratio, and centering.
- The xref integrity test follows `startxref` and validates the declared object rows.

I locally merged this branch on top of current `main`; it merges cleanly and `cargo test --locked` passes (203/203), `python3 scripts/check_i18n.py` passes, and `git diff --check` passes.

One remaining hygiene issue before approval/merge: `cargo fmt --check` reports formatting diffs in this PR's changed files (`src/app_state.rs`, `src/pdf.rs`, `src/ui/layout_image_export.rs`) in addition to the repository's existing old baseline. Please run/apply rustfmt for the new changes and rerun CI/checks.

ImmortalDragonm:
Applied rustfmt to this PR's changed files (`src/app_state.rs`, `src/pdf.rs`, `src/ui/layout_image_export.rs`) so `cargo fmt --check` passes on the new code (leaving the repo's existing baseline untouched). Tests still 145/145.

src/pdf.rs
```diff
@@ -0,0 +1,299 @@
...
+            "{image_id} 0 obj\n<< /Type /XObject /Subtype /Image \
+             /Width {} /Height {} /ColorSpace /DeviceRGB /BitsPerComponent 8 \
+             /Filter /FlateDecode /Length {} >>\nstream\n",
+            image.width(),
+            image.height(),
+            data.len()
+        )
+        .as_bytes(),
+    );
+    pdf.extend_from_slice(&data);
```
IgorArkhipov, line 145: Nice! Rendering is now one page at a time, and latest formatting cleanup looks good.

One resource-bound edge remains: encode_rgb still holds page RGBA + full RGB + compressed bytes, while this copy grows an in-memory PDF containing all prior pages.

Current <12,000 px side guard can admit an 11,999 x 11,999 page, roughly 1 GB in raw RGBA + RGB buffers before compressed/output storage, and 32 pages have no aggregate bound.

Could we possibly add checked per-page/total pixel limits?
Feeding RGB rows directly into ZlibEncoder would reduce peak further.

ImmortalDragonm:
Addressed the resource-bound point: `encode_rgb` now streams one row at a time into the ZlibEncoder (no full-image RGB copy), and `build_layer_pdf` enforces a per-page (~30 MP) and whole-document (~120 MP) pixel budget via `accumulate_pixel_budget`, failing cleanly instead of exhausting memory. Added tests for the accumulation and both rejection paths.

kissetfall (changes requested):
Thanks — the overall PDF export structure and UI integration look good, but one resource-safety issue blocks merge.

`build_layer_pdf` calls `render_page(page)` before `accumulate_pixel_budget`. The renderer has already created the `RgbaImage` by then, so the advertised 30 MP per-page guard cannot prevent the large allocation it is meant to protect against. The export geometry currently permits dimensions close to 12,000 × 12,000, which can allocate roughly 576 MiB of RGBA memory before the PDF code rejects the page.

Please move the per-page pixel preflight ahead of image allocation/rendering. This may require passing page dimensions into the PDF builder or validating the export geometry before invoking the render closure. Add a regression test proving that an oversized page is rejected before the renderer/allocation is invoked. The total-document budget can remain in the PDF writer.

Ready for re-review once the pre-allocation guard and regression test are in place. The branch otherwise merges cleanly with current `main`; the merged test suite, clippy, and release build passed locally.

kissetfall (approved):
Re-reviewed head c06ba7a. The page pixel budget is now checked from precomputed geometry before render_layout_image can allocate an RgbaImage, and the regression proves that an oversized page never invokes the renderer. The streaming one-page-at-a-time and total-document limits remain intact.

Local verification passed 149/149 tests, scoped rustfmt, i18n, diff-check, Clippy, and a release build. GitHub has no hosted checks configured for this branch, but the PR is mergeable and the requested resource-safety fix is present.

Approved.

PR #87, opened 2026-07-15
Invalidate stale QMK settings cache

IgorArkhipov:
### Problem

Entropy reused the cached supported-QSID list after firmware or Vial definition changes. Controller fields could stay hidden or appear unsupported until users manually removed application cache files.

### Fix

- Version QMK settings cache entries with Vial definition size and fingerprint, VIA/Vial protocol versions, and firmware version.
- Re-query supported QMK settings whenever that metadata changes; legacy unversioned cache entries are rejected and replaced.
- Add **Refresh device data** under About Device. It removes every cached definition size plus QMK settings for only the selected device, then reconnects.

Matching cache metadata still keeps normal reconnects fast.

### Verification

- `cargo test --locked`: 196 passed.
- `cargo clippy --locked --all-targets`: passed with existing repository warnings.
- `python3 scripts/check_i18n.py`: passed.
- Scoped `rustfmt --check` and `git diff --check`: passed.
- `TARGET=aarch64-apple-darwin scripts/build_macos_app.sh`: arm64 app, DMG, and ZIP built; plist validation passed.

Repository-wide `cargo fmt -- --check` still reports pre-existing formatting baseline outside this change.

### Проблема

Entropy продолжала использовать сохраненный список поддерживаемых QSID после изменения прошивки или Vial definition. Настройки контроллеров могли оставаться скрытыми или выглядеть неподдерживаемыми, пока пользователь вручную не удалял cache-файлы приложения.

### Исправление

- Записи cache для QMK settings теперь версионируются вместе с размером и fingerprint Vial definition, версиями протоколов VIA/Vial и версией прошивки.
- При изменении этих метаданных список поддерживаемых QMK settings запрашивается заново; старые записи без метаданных отклоняются и заменяются.
- На странице About Device добавлена кнопка **Обновить данные**. Она удаляет все сохраненные размеры definition и QMK settings только для выбранного устройства, затем переподключается.

При совпадении метаданных обычное переподключение по-прежнему использует быстрый cache path.

### Проверка

- `cargo test --locked`: прошли 196 тестов.
- `cargo clippy --locked --all-targets`: прошел с существующими предупреждениями репозитория.
- `python3 scripts/check_i18n.py`: прошел.
- Scoped `rustfmt --check` и `git diff --check`: прошли.
- `TARGET=aarch64-apple-darwin scripts/build_macos_app.sh`: собраны arm64-приложение, DMG и ZIP; plist прошел проверку.

Общий `cargo fmt -- --check` по репозиторию по-прежнему сообщает о ранее существовавшем formatting baseline вне этого изменения.

### DIFF
--- a/i18n/en.toml
+++ b/i18n/en.toml
@@ -443,6 +443,10 @@ device_unlocked = "Device unlocked!"
 unlock_interrupted_disconnected = "Unlock interrupted — device disconnected or unavailable: {error}"
 unlock_poll_retry = "Unlock communication interrupted; retrying: {error}"
 finish_unlock_before_closing = "Finish device unlock or reconnect it before closing Entropy"
+refresh_device_data_pending_write = "Finish the pending layer write before refreshing device data"
+refresh_device_data_missing_device = "Failed to refresh device data: device not found"
+refresh_device_data_missing_info = "Failed to refresh device data: device information unavailable"
+refresh_device_data_delete_failed = "Failed to remove cached device data"
 
 [dynamic_status]
 device_not_found = "Device not found"
--- a/i18n/ru.toml
+++ b/i18n/ru.toml
@@ -443,6 +443,10 @@ device_unlocked = "Устройство разблокировано"
 unlock_interrupted_disconnected = "Разблокировка прервана — устройство отключено или недоступно: {error}"
 unlock_poll_retry = "Связь при разблокировке прервана; повторная попытка: {error}"
 finish_unlock_before_closing = "Завершите разблокировку устройства или переподключите его перед закрытием Entropy"
+refresh_device_data_pending_write = "Завершите запись слоя перед обновлением данных устройства"
+refresh_device_data_missing_device = "Не удалось обновить данные устройства: устройство не найдено"
+refresh_device_data_missing_info = "Не удалось обновить данные устройства: информация недоступна"
+refresh_device_data_delete_failed = "Не удалось удалить сохранённые данные устройства"
 
 [dynamic_status]
 device_not_found = "Устройство не найдено"
--- a/src/ui/about_device.rs
+++ b/src/ui/about_device.rs
@@ -522,14 +522,10 @@ fn draw_about_rows(
     id_salt: &'static str,
     metrics: crate::ui_style::ResponsiveMetrics,
     rows: &[AboutRow],
+    bottom_reserve: f32,
 ) -> egui::Rect {
-    let list = allocate_adaptive_settings_list_viewport(
-        ui,
-        id_salt,
-        metrics,
-        rows.len(),
-        metrics.value(4.0),
-    );
+    let list =
+        allocate_adaptive_settings_list_viewport(ui, id_salt, metrics, rows.len(), bottom_reserve);
 
     crate::ui_style::allocate_ui_at_rect(ui, list.content_rect, |ui| {
         ui.set_clip_rect(list.viewport);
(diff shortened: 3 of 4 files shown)


kissetfall (changes requested):
Blocking correctness issue: the new definition_fingerprint is computed from json only after load_cached_vial_definition has already selected a cached Vial definition by cache_key and definition_size. If the device definition changes while its compressed size stays the same, the application never reads the new definition; it fingerprints the old cached JSON and can therefore reuse the old QMK settings cache. The fallback firmware_version_from_vial_json is stale in the same path, so devices without a runtime firmware version remain affected. The current same-size unit test bypasses the actual cache-selection path by directly passing two different JSON values.

Please make the reuse decision depend on freshness metadata read directly from the device, or disable QMK cache reuse when no trustworthy runtime freshness token exists, and add a test covering the real stale-definition cache path. The scoped tests, rustfmt, i18n check, diff check, and clippy all pass; this is a logic blocker rather than a build failure.

IgorArkhipov:
Thanks, you're right.

I'll make QMK settings-cache reuse depend on freshness metadata read directly from the device; when no trustworthy runtime freshness token is available, reuse will be disabled.

Will update the PR with a fresh CI run once that is in place.

IgorArkhipov:
Addressed feedback in commit 1d00de7.

`device_connect_task.rs` now reads runtime firmware version before cache selection.
Cached Vial definitions and QMK settings reuse only when that direct device token matches; JSON fallback version cannot authorize reuse.

Devices without a usable runtime version reload definition and re-query QMK settings.

Added on-disk regression coverage for same-size definitions with changed firmware version.

IgorArkhipov:
Tested full flow using macros and assigning symbols to them.
First revoked all permissions, closed the app, reopened, requested permissions and then granted both Accessibility and Input Monitoring under Settings / Privacy & Security

kissetfall (changes requested):
The original stale-definition blocker is fixed: runtime firmware metadata is now read before cache selection, cache reuse is disabled without a trustworthy direct token, and the on-disk same-size regression path is covered. The updated branch also merges cleanly with current `main`; 221 tests, clippy, i18n, scoped rustfmt, and diff checks pass locally.

Three blockers remain:

1. `draw_about_rows` reserves only 4 px below the 20/22-row About Device list, while the new **Refresh device data** action extends about 50 px below the viewport. On constrained window sizes the fixed action can be clipped outside the page. Please pass an action-aware bottom reserve through the shared list allocation, following the existing settings pages that reserve space for fixed buttons.

2. `is_cached_vial_definition_for_device` uses `starts_with`, so a cache key such as `keyboard` also matches `definition_v4_keyboard_serial_00001234.json`. Refreshing one device can therefore remove another device's definition cache when one cache key is a prefix of the other. Please parse the final `_XXXXXXXX.json` size suffix and require an exact cache-key match, with a regression test for prefix-related keys.

3. The user-visible failure messages in `refresh_current_device_data` are hardcoded in English, including the pending-write, missing-device, missing-info, and deletion-error paths. Please move this copy into the translation catalogs and select it through the current application language.

After these changes I will re-review the new delta against current `main`.

kissetfall (approved):
Re-reviewed head e1b9fa3 against current main. All three requested changes are resolved: fixed-action space is reserved, definition cache deletion requires an exact device key with prefix-key regression coverage, and refresh errors are localized. The cache and reconnect lifecycle are consistent with the existing owners.

Local validation passed: 221/221 tests, Clippy, i18n, scoped rustfmt, git diff --check, and release build. CI is 4/4 green. No remaining blocker.

PR #97, opened 2026-07-20
feat(macros): add host-assisted Unicode text

IgorArkhipov:
Addresses https://t.me/c/1464748383/37027/131938 report.

Summary:
- Add a Unicode macro action that transports UTF-8 through standard Vial key actions and inserts it on macOS, Windows, X11, IBus, and Fcitx5.
- Keep existing Text macros unchanged; enforce UI and device macro-buffer limits.
- Reset interrupted transports safely while preserving modifier transitions required by a valid macro.

Validation:
- cargo test: 331 passed.
- Python syntax check for linux/ibus/entropy-ibus-engine passed.
- git diff --check passed.

Post-merge validation:
- Assign a Unicode macro such as ß, save it, and trigger it in a native text field on each supported backend.
- Healthy: exact text is inserted once; interrupted transports emit nothing.
- Failure: disable Universal Symbols or Host Text macro use and investigate backend logs.


kissetfall:
Thank you for the implementation and the cross-platform work behind it. After reviewing this together with the related host-assisted text proposals, we have decided not to pursue this feature direction and will not merge the PR.

The Unicode macro depends on Entropy or an input-method backend being continuously available. Without that backend, the keyboard emits a long transport sequence built from modifiers and F13-F20, which may be interpreted by the operating system or other applications as real shortcuts. The current character limit also does not represent the actual encoded macro-buffer cost, especially for multi-byte emoji, and the claimed platform behavior has not been validated with hardware-level HID timing tests.

These constraints make the feature less portable and predictable than a keyboard macro should be, while adding a substantial host-side runtime dependency. This is the same product-level reason we declined the related host-text implementation, so we are closing this PR without merge. We can reconsider Unicode input in the future if it has safe behavior when Entropy is absent, buffer-aware validation, and verified end-to-end behavior on the supported platforms.

PR #101, opened 2026-07-24
Preserve portable firmware settings in .entlayout

IgorArkhipov:
## Summary

This is the focused follow-up suggested in the closing comment on #93.

It extends manual `.entlayout` export/import without adding automatic recovery infrastructure.

- Bump `.entlayout` to v4 and add an optional `portable_settings` section for advertised standard QMK settings, including Quick Tap, Auto Shift, One Shot, Tap-Hold, Mouse Keys, Magic, Grave Escape, and compatible touchpad settings such as Auto Layer.
- Store a stable semantic ID, QSID, value width, value, and—where applicable—the ordered variants that define a select field's meaning.
- Restore only settings whose target contract matches. Unsupported, malformed, duplicated, or incompatible entries are skipped conservatively.
- Verify every write with immediate exact readback, continue after individual failures, and show imported/unsupported/incompatible/failed counts plus exact write failures in the import report.
- Preserve v1-v3 import compatibility; files without `portable_settings` continue to load unchanged.

## Scope

Existing `.entlayout` sections remain authoritative for layer names and Combo Term, so they are not duplicated. Keyboard-specific module fields, Layer LED/RGB, and Bluetooth settings are intentionally excluded because their portable contract cannot be established generically across firmware definitions.

This PR does not add serial-backed history, automatic prompts, deferred banners, background capture, or a separate HID recovery lifecycle.

## Validation

- `cargo test --locked`: 320 tests passed
- `cargo build --release --locked`: passed
- `cargo clippy --locked --bin entropy`: passed (repository's existing warnings only)
- `python3 scripts/check_i18n.py`: passed
- `rustfmt --edition 2021 --check src/entlayout.rs`: passed
- `git diff --check`: passed


kissetfall:
Updated the branch on top of current main and addressed the review blockers:

- bumped .entlayout to v5 while retaining v1-v4 imports;
- moved the portable QMK settings contract/state mapping into its own owner module;
- made staged Bluetooth export/import load Behavior Settings and Touchpad first;
- export only values confirmed by successful reads, including independent One Shot QSIDs;
- skip settings already equal on the target;
- reuse the existing settings writer policy: ordinary settings are SET-only, touchpad settings use delayed verified readback;
- added coverage for deferred BLE data, failed reads, unchanged values, format compatibility, and write/readback policy.

Local checks: 497/497 tests passed and the i18n catalog check passed. CI is now running on the updated head.

PR #106, opened 2026-07-28
Fail device loads when HID disconnects

IgorArkhipov:
### Problem

When the mandatory initial USB keymap read failed, Entropy logged the error and
continued with zero-filled layers. During deferred loading, a real disconnect
while reading encoders or firmware layer names was also treated like an
unsupported optional feature, so a lost device could appear successfully
loaded with defaults.

### Fix

- Mandatory initial keymap reads now fail the USB or staged Bluetooth
  connection instead of accepting zero-filled layers.
- Deferred encoder and firmware-layer-name reads now propagate recognized
  disconnects through the existing reconnect path.
- Non-disconnect errors for optional features retain the existing default
  fallback behavior.
- Fault-injection tests verify the exact HID command that failed, covering both
  direct and background layer loading.

### Verification

- `cargo test device_deferred_load` - 29 passed.
- `cargo test device_connect_task` - 23 passed.
- `cargo test -- --test-threads=1` - 444 passed.
- Scoped `rustfmt --check` for both changed files passed.
- `git diff --check` passed.
- The ordinary parallel suite still has two pre-existing timing-sensitive
  failures in `vial_hid_task`; the untouched `origin/main` worktree reproduces
  them, and the serial full suite is green.

### DIFF
--- a/src/ui/device_connect_task.rs
+++ b/src/ui/device_connect_task.rs
@@ -323,6 +323,24 @@ fn normalize_reported_layer_count(reported_layer_count: usize) -> usize {
     reported_layer_count.max(1)
 }
 
+#[cfg(not(target_arch = "wasm32"))]
+fn read_initial_keymap_buffer(
+    hid: &crate::hid::HidDevice,
+    layer_count: usize,
+    rows: usize,
+    cols: usize,
+    staged_bluetooth_load: bool,
+) -> Result<Vec<u16>, String> {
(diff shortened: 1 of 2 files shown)


kissetfall:
Thanks for the contribution. The mandatory initial keymap failure handling is useful, but this implementation also treats transient timeouts and generic HID I/O failures as disconnects for optional deferred reads. That can trigger unnecessary reconnects, so we've decided not to proceed with this PR and are closing it without merging.

PR #109, opened 2026-07-28
Keep Matrix Tester HID reads off the UI thread

IgorArkhipov:
### Problem

Matrix Tester polled USB keyboards directly from the UI thread at a 16 ms cadence. A slow HID round-trip could therefore make the app hitch while the tester was open, even though Bluetooth matrix reads already used the serialized background worker.

Moving all matrix reads to that worker also required a narrower interaction policy: ordinary editing should stay available during read-only polling, while HID-backed undo, whole-layer writes, and unlocked-device lock actions must still wait for the current read to release the device.

### Fix

- USB and Bluetooth matrix reads now run through the shared background Vial HID worker, with HID ownership restored before the next operation.
- Matrix and deferred background reads no longer disable ordinary editing; undo and whole-layer actions queue and run after the read completes.
- Lock actions remain disabled when an unlocked keyboard has an active unqueued read.
- Matrix polling avoids redundant repaint wakeups while work is active and does not repaint when no device is available.
- Regression tests cover the USB matrix command, HID restoration, nonblocking editing, queued undo and layer writes, and deadline-based async test waits.

### Verification

- `cargo test matrix_tester --locked` - 4 passed
- `cargo test vial_hid_task --locked` - 12 passed
- `cargo test --locked` - 438 passed with the normal parallel runner
- `cargo clippy --locked --all-targets` - passed with existing warnings
- `python3 scripts/check_i18n.py`
- Scoped `rustfmt --edition 2021 --check` for all five changed Rust files
- `git diff --check origin/main`

### DIFF
--- a/src/ui/key_assignment.rs
+++ b/src/ui/key_assignment.rs
@@ -377,9 +377,9 @@ impl EntropyApp {
 
     #[cfg(not(target_arch = "wasm32"))]
     pub(super) fn undo(&mut self, ctx: &egui::Context) {
-        if self.vial_hid_background_layer_active() {
+        if self.vial_hid_nonblocking_read_active() {
             self.pending_layout_undo = true;
-            self.deferred_device_load.defer_background_for_user_input();
+            self.defer_background_layer_for_user_input_if_active();
             ctx.request_repaint_after(std::time::Duration::from_millis(16));
             return;
         }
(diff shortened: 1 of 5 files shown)


kissetfall:
Thanks for the contribution. We've decided not to implement this change and will keep the current Matrix Tester polling and interaction behavior. Closing without merging.

PR #111, opened 2026-07-29
Verify AppImage packaging tool downloads

IgorArkhipov:
### Problem

Linux release packaging downloaded and executed AppImageKit's `appimagetool` without verifying its contents.
A corrupted or replaced upstream download, or a stale cached executable, could therefore be trusted as release tooling and alter the produced AppImage.

The upstream x86_64 tool is published under a rolling `continuous` asset, so a fixed digest also needs to be exercised before a release tag rather than failing for the first time during release packaging.

### Fix

- The current AppImageKit asset URL and SHA-256 digest now share one reviewed pin.
- Fresh downloads are staged in a temporary file, verified before becoming executable, and moved into the cache only after validation.
- Existing cached tools are re-verified before every use.
- Linux CI verifies the live pinned asset and exercises valid and corrupt fresh-download and cached-tool paths, including cleanup and proof that mismatched tools never execute.

### Verification

- `bash scripts/test_verify_sha256.sh`
- `bash scripts/test_build_linux_appimage.sh`
- `bash -n scripts/appimagetool_pin.sh scripts/build_linux_appimage.sh scripts/verify_sha256.sh scripts/test_verify_sha256.sh scripts/test_build_linux_appimage.sh`
- `shellcheck scripts/appimagetool_pin.sh scripts/build_linux_appimage.sh scripts/verify_sha256.sh scripts/test_verify_sha256.sh scripts/test_build_linux_appimage.sh`
- `actionlint .github/workflows/build.yml`
- Ruby YAML parse of `.github/workflows/build.yml`
- Live AppImageKit asset matched the pinned SHA-256 digest
- `git diff --check origin/main...HEAD`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,6 +61,10 @@ jobs:
             libssl-dev \
             libgtk-3-dev
 
+      - name: Test AppImage tool integrity verification
+        if: matrix.kind == 'linux'
+        run: bash scripts/test_verify_sha256.sh
+
       - name: Build
         if: matrix.kind != 'macos'
         run: cargo build --release
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,7 +63,9 @@ jobs:
 
       - name: Test AppImage tool integrity verification
         if: matrix.kind == 'linux'
-        run: bash scripts/test_verify_sha256.sh
+        run: |
+          bash scripts/test_verify_sha256.sh
+          bash scripts/test_build_linux_appimage.sh
 
       - name: Build
         if: matrix.kind != 'macos'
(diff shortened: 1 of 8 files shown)


kissetfall:
Keeping this open because checksum verification is useful. One blocker remains: the URL points to the mutable AppImageKit continuous asset while the checksum is fixed, so a future upstream rebuild will break packaging. Please pin an immutable release/tag asset that matches the checksum, or add an explicit update mechanism.

IgorArkhipov:
Addressed in `4988ac7`.

AppImageTool is now pinned to immutable AppImageKit release `12` with its matching SHA-256.
I also added `scripts/update_appimagetool_pin.sh <release-tag>` to download an immutable release asset, recompute its checksum, and atomically update the pin; it rejects `continuous` and unsafe tags.

The updater has regression coverage and is included in the Linux AppImage integrity test step.
The focused shell and AppImage integration tests pass locally.

PR #112, opened 2026-07-29
Avoid repeated combo highlight resolution during repaints

IgorArkhipov:
### Problem

Combo highlighting resolved every displayed key through transparent-layer fallbacks separately for every configured combo on every keyboard repaint. The resolved keycode is identical across combos, so this repeated work scaled with combos × keys × fallback depth.

A temporary representative release-mode probe measured the one-resolution-per-key approach before the final review removed its scratch allocation:

| Highlight pass | Mean per repaint |
| --- | ---: |
| `origin/main` | 21.606 µs |
| One resolution per key | 4.606 µs |

### Fix

- Combo highlighting now traverses physical keys once, resolves each effective keycode once, and checks configured combos in their original order.
- The repaint path adds no cache, invalidation state, or resolved-keycode allocation.
- Incomplete combo entries return empty highlights without layer fallback work.
- Characterization tests preserve transparent fallback, conflict ordering and colors, uncolored membership, incomplete entries, zero keycodes, and empty combo behavior.

### Verification

- `cargo test combo_highlights_` - 8 passed
- `cargo test -- --test-threads=1` - 445 passed
- Scoped `rustfmt --edition 2021 src/ui/layout_keyboard.rs`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/src/ui/layout_keyboard.rs
+++ b/src/ui/layout_keyboard.rs
@@ -857,6 +857,10 @@ impl EntropyApp {
     }
 }
 
+fn combo_paints_highlights(combo: &ComboEntry) -> bool {
+    combo.output != 0 && combo.keys.iter().any(|&key| key != 0)
+}
+
 fn combo_key_colors_for_layer(
     layout: &KeyboardLayout,
     layer: usize,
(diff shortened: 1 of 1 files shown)


kissetfall:
Thanks for the contribution. We decided not to pursue this optimization and will keep the current Combo highlighting path. Closing without merge.

PR #115, opened 2026-07-29
Add Nix flake with NixOS and home-manager modules

AlexBSoD:
Adds a Nix flake so Entropy can be installed and developed on NixOS.

Two of the Linux setup steps documented in the README cannot work there, and the flake handles both declaratively:

- **Vial udev rules.** `linux/udev/install-vial-rules.sh` writes `/etc/udev/rules.d/59-vial.rules`, which NixOS discards on the next activation. The NixOS module emits the same two rules through `services.udev`.
- **IBus engine.** `linux/ibus/install-user.sh` installs into `~/.local/share/ibus/component`, but on NixOS `ibus-with-plugins` pins `IBUS_COMPONENT_PATH` to its own store path with `--set`, so nothing outside it is ever loaded. The module registers the engine through `i18n.inputMethod.ibus.engines`, which is the only path IBus honours.

The engine is a separate package: the bundled python3 script is wrapped with PyGObject and the IBus typelib, because its `#!/usr/bin/env python3` shebang would otherwise resolve to an interpreter without the `gi` bindings. Its `--entropy-check` passes.

### What's included

| Output | |
| --- | --- |
| `packages.<system>.entropy` | the app, with desktop entry and icon |
| `packages.<system>.entropy-ibus-engine` | wrapped engine + IBus component XML |
| `nixosModules.default` | app, udev rule, IBus engine |
| `homeManagerModules.default` | app and IBus engine (no udev, needs root) |
| `devShells.<system>.default` | Rust toolchain and native dependencies |

`x86_64-linux` and `aarch64-linux`. Usage is documented in a new `## NixOS` section in the README.

### Packaging notes

A few things that differ from what the CI installs, in case they are useful beyond this PR:

- `hidapi` is built with the `linux-native-basic-udev` feature, which resolves to the pure-Rust `basic-udev` crate — neither `libudev` nor `libhidapi` is actually needed to build.
- `gtk3` is needed by `rfd` for file dialogs, not by the app itself (`gtk-sys` appears in `Cargo.lock`).
- TLS comes from rustls via `ureq`, so `libssl` is unused.
- X11, Wayland, GL and libxkbcommon are dlopened at runtime, so they are passed through `LD_LIBRARY_PATH` in the wrapper rather than through `buildInputs` alone.
- The repository ships no Linux icon or desktop entry, so the flake generates both, reusing the SVG artwork from `scripts/build_linux_appimage.sh`.

### Verification

- `nix build .#entropy .#entropy-ibus-engine` — both build, `cargo test` runs as part of the package build and passes
- the built binary starts and `ldd` reports no missing libraries
- `pkgs.ibus-with-plugins.override { plugins = [ engine ]; }` builds, i.e. the engine is accepted as an IBus engine
- evaluating a throwaway NixOS configuration with `programs.entropy.enable = true` yields the expected udev rules and engine registration, with no warnings
- `nix flake check` passes

### Two decisions worth a second opinion

1. `programs.entropy.ibus.enable` defaults to `true` and sets `i18n.inputMethod.enable`/`type` via `mkDefault`, so enabling Entropy turns IBus on for the whole system. That may be more than you want as a default — happy to flip it to `false`.
2. The lock pins `nixos-unstable`, and the flake uses the current package names (`libx11` rather than `xorg.libX11`), so it will not evaluate on 25.05. If stable support matters, the old attribute paths can be restored.

`flake.lock` is committed, as is usual for an application flake, so builds are reproducible. It will need an occasional `nix flake update`.


nix/hm-module.nix
```diff
@@ -0,0 +1,72 @@
...
+  config = lib.mkIf cfg.enable (
+    lib.mkMerge [
+      { home.packages = [ cfg.package ]; }
+
+      (lib.mkIf cfg.ibus.enable {
+        home.packages = [ cfg.ibus.package ];
+
+        # IBus reads components from this list only; keep the distro default so
+        # the other engines do not disappear.
+        home.sessionVariables.IBUS_COMPONENT_PATH = "${cfg.ibus.package}/share/ibus/component:${pkgs.ibus}/share/ibus/component";
```
IgorArkhipov: This replaces `IBUS_COMPONENT_PATH` with only Entropy and the base IBus directories, which can hide other engines installed in the user profile.

Could we prepend Entropy's component directory while preserving the existing path?

kissetfall:
The core Nix packaging is still relevant, so I am keeping this PR open. Please update it against current Entropy: Universal Symbols are now native RMK actions gated by firmware capability; IBus remains only for Text Expander. The README and module options still describe IBus as serving Universal Symbols, which is now stale.

AlexBSoD:
Updated and rebased onto `1ce74b5`.

* Both module options and the README section no longer describe the engine as serving Universal Symbols — it is now documented as backing Text Expander on Wayland, with a note that Universal Symbols need no input method on firmware exposing native RMK key actions.
* Fixed the input source name while I was there: the component registers **Entropy Symbols** (plus the layout variants **Entropy EN**, **Entropy RU**, …), not "Entropy Universal Symbols".
* Added a line that `programs.entropy.ibus.enable = false;` is enough when Text Expander is not used.

`nix build .#entropy .#entropy-ibus-engine`, a NixOS-module eval and `nix fmt` are all clean on the rebased branch.

One thing I could not confirm from the repository, so please correct me if I am looking in the wrong place: on `1ce74b5` the app still routes Universal Symbols through the host rather than through native RMK actions. The picker writes `F13`–`F24` carriers (`smart_input_symbols.rs`, `keycode_picker.rs:382`), `linux_recommended_input_backend()` returns `IBus` for Wayland (`smart_input.rs:210`), the setup wizard shells out to `linux/ibus/install-user.sh` (`ui/universal_symbols_setup.rs`), and `rmk_native.rs` only carries `TapHold` / `Modifier` / `KeyWithModifier`. If the native path lands in a later release, the wording here is already written for it — if it is expected to work on `main` today, point me at it and I will adjust.

Related: the upstream "Linux IBus Backend" section (README, above the NixOS one) still says IBus serves Universal Symbols. I left it untouched to keep this PR scoped to the Nix bits, but I am happy to update it here if you would rather have both in one go.

nix/hm-module.nix
AlexBSoD: Addressed — the module now prepends through `home.sessionSearchVariables.IBUS_COMPONENT_PATH`, so Entropy and the base IBus directories go in front and whatever was already on the path is kept. Thanks for catching it.

AlexBSoD:
Rebased onto `92f0969` (v0.3.2-rc.1). Nothing in the flake needed adjusting for it: the only manifest change was the version bump, both packages build (`nix build .#entropy .#entropy-ibus-engine`), and the NixOS module still evaluates clean.

My earlier question is moot — `smart_input_symbols.rs` is gone and `universal_symbols.rs` plus the native RMK actions are on `main` now, which matches the documentation wording already in this branch.

One new thing worth flagging: nixpkgs has picked up Entropy on its own — `pkgs.ergohaven-entropy` (0.3.1) plus `nixos/modules/programs/ergohaven-entropy.nix`, maintained by @geekiot-hub. That module declares the same `programs.entropy.enable` option, so importing this flake's module alongside a current nixpkgs aborted evaluation with *"The option `programs.entropy.enable` ... is already declared"*. I've handled it with `disabledModules = [ "programs/ergohaven-entropy.nix" ]` so this flake's module stays authoritative — it tracks the version in this repo rather than the last released tag, and it lets the package be overridden.

Separately, and independently of this PR: the nixpkgs module is currently broken. `pkgs.ergohaven-entropy` does not set `meta.isIbusEngine = true`, so `programs.entropy.enable = true` on a plain unstable checkout fails with

```
error: A definition for option `i18n.inputMethod.ibus.engines."[definition 1-entry 1]"' is not of type `ibus-engine'
```

Happy to send that one-line fix, along with a 0.3.2 bump, to nixpkgs if that is useful.

kissetfall (changes requested):
Thanks for the contribution. The core Nix packaging is useful, but the following changes are required before merge:

1. Make the IBus integration opt-in in both modules (default false). Enabling Entropy alone must not enable or select a system-wide input method.
2. Do not set the system-wide input method merely because the app is enabled. Require an explicit IBus choice, and warn or document when the active input method is not IBus.
3. Align the integration and documentation with the current Text Expander implementation. Entropy should recognize the declaratively installed engine instead of telling users to ignore an incorrect installation state; the documented input-source names must match the actual Entropy Text Expander names; and the stale Universal Symbols package description should be removed.
4. Add automated Nix coverage: nix flake check, builds of both packages, and a NixOS module evaluation. The existing Build workflow does not validate the flake.
5. Clarify and test coexistence with the Entropy module now shipped by nixpkgs. If this module intentionally supersedes it, keep that behavior explicit and covered so we do not maintain two silently conflicting sources of truth.

Please rebase onto current main after addressing these points.

AlexBSoD:
Thanks — all five points are addressed, rebased onto `b9fa8c3`.

**1. IBus is opt-in.** `programs.entropy.ibus.enable` now defaults to `false` in both the NixOS and the home-manager module. `programs.entropy.enable` on its own installs the app and the Vial udev rule, nothing else.

**2. No input method is chosen for you.** The module no longer sets `i18n.inputMethod.enable` or `i18n.inputMethod.type`, not even with `mkDefault`. With `ibus.enable = true` it only appends the engine to `i18n.inputMethod.ibus.engines`, and warns when the active input method is disabled or is not IBus — the case where the engine would silently never load. The README shows the explicit `i18n.inputMethod` block next to it.

**3. Aligned with the current Text Expander implementation.** Entropy now recognises an engine registered outside `$XDG_DATA_HOME`: it resolves the component the way IBus does (`IBUS_COMPONENT_PATH` first, then every `XDG_DATA_DIRS` entry) and, when the registration is not user-local, replaces the **Install IBus** / **Uninstall IBus** actions with a note instead of offering an action that would only add an unmanaged copy. Five unit tests cover missing / user / system / both / unrelated-component cases. The documented input sources are now the real names — **Entropy Text Expander** and its layout variants — and the engine package no longer claims to serve Universal Symbols.

**4. Automated Nix coverage.** New `.github/workflows/nix.yml` on pull requests and pushes to `main`: `nix fmt -- --ci`, the module evaluation, `nix build` of both packages, the engine's own `--entropy-check`, and `nix flake check`. It needs your "Approve and run" — the run currently sits at `action_required`, as does the Build workflow.

**5. Coexistence with the nixpkgs module, explicit and tested.** This module deliberately supersedes `nixos/modules/programs/ergohaven-entropy.nix` via `disabledModules`, which is now stated in the README as a choice made by importing it. `nix/checks.nix` evaluates the module on a throwaway machine, and since `nixosSystem` pulls in all of nixpkgs' modules, every assertion also proves the two do not collide: it checks that the app and the rule are installed, that no input method is touched, that `ibus.enable` registers exactly the Entropy engine and warns otherwise, that `programs.entropy.ibus` still exists (i.e. our module won), and that `programs.entropy.package` is this flake's package rather than `pkgs.ergohaven-entropy`.

Local run: `cargo test --all-targets` 524 passed, `nix flake check` clean, both packages build, `actionlint` clean on the new workflow.

One thing you may want to know regardless of this PR: the nixpkgs module is broken today. `pkgs.ergohaven-entropy` does not set `meta.isIbusEngine = true`, so `programs.entropy.enable = true` on plain unstable fails with `A definition for option i18n.inputMethod.ibus.engines."[definition 1-entry 1]" is not of type ibus-engine`. Happy to send that one-line fix and a 0.3.2 bump there if you would rather keep the nixpkgs path working.

nix/module.nix
```diff
@@ -0,0 +1,130 @@
...
+  options.programs.entropy = {
+    enable = lib.mkEnableOption "Entropy keyboard configurator";
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      defaultText = lib.literalExpression "entropy.packages.\${system}.entropy";
+      description = "The Entropy package to use.";
+    };
+
+    group = lib.mkOption {
```
IgorArkhipov, line 60: Thanks for making the access group configurable!

Could we either create the selected group or explicitly validate/document that it must already exist?

A custom value can evaluate and reach the udev rule without a corresponding group, leaving the hidraw device inaccessible.

nix/hm-module.nix
```diff
@@ -0,0 +1,80 @@
...
+
+        # IBus reads components from this list only; prepend our paths and
+        # keep whatever IBUS_COMPONENT_PATH already had, so other engines
+        # relying on it do not disappear.
+        home.sessionSearchVariables.IBUS_COMPONENT_PATH = [
+          "${cfg.ibus.package}/share/ibus/component"
+          "${pkgs.ibus}/share/ibus/component"
+        ];
+
+        # IBus caches the registry — after switching generations run
```
IgorArkhipov: One lifecycle gap: after a Home Manager switch, Entropy sees the new engine and hides its recovery actions, while the running IBus daemon can still use its old registry.

Could we surface the required IBus restart in the user-facing guidance, or keep a safe recovery path until it reloads?

nix/module.nix
AlexBSoD: Now created by the module: `users.groups.${cfg.group} = { }`. A no-op for the default `input` (its gid 174 is kept) and works for any custom value; covered in `nix/checks.nix`.

nix/hm-module.nix
AlexBSoD: Good catch — hiding the install actions closed the only recovery path.

Entropy now shows **Reload IBus registry** in their place when the engine is registered outside `$XDG_DATA_HOME`: it reruns `ibus write-cache` and restarts the daemon, writing nothing. The README and the module comment point at it.

AlexBSoD:
Rebased onto `613fdb4`; both review notes are answered in their threads. The udev group is now created by the module, and a system-registered engine gets a **Reload IBus registry** action instead of a dead end.

AlexBSoD:
Self-review caught three problems in my own detection code, now fixed: it scanned the component directories on every frame, a user install could be misread as a system one (XDG_DATA_DIRS often contains paths under $HOME), and a user copy next to a system registration lost its uninstall action. User and system presence are tracked separately now, the result is cached, and the daemon reload runs off the UI thread.

Also: the group check no longer hardcodes gid 174, and the CI build step names its outputs instead of relying on `result-1`.

Sorry for the number of pushes on this one — embarrassing for what started as a packaging PR.

kissetfall (changes requested):
Thanks — the five points from the previous review appear to be addressed. One runtime-state issue remains before merge.

`EntropyApp::ibus_registration` caches the first component-directory scan indefinitely. A NixOS/home-manager rebuild can add or remove the system component while Entropy is still running; if the setup page was opened before that rebuild, the cached `system: false` value remains and the documented **Reload IBus registry** action never appears until the app is restarted or an in-app setup action happens.

Please make externally managed registration changes observable without restarting Entropy (for example, rescan on page entry/focus, use a bounded refresh interval, or invalidate from a cheap directory fingerprint) and add regression coverage for an external registration change. Please also route the two new reload failure prefixes/statuses through i18n instead of showing hard-coded English in the Russian UI.

After that, please rebase the remaining one commit onto current `main`. A synthetic merge passed 570/570 Rust tests, i18n, and Clippy locally, but both GitHub workflows are still `action_required`; the Build and Nix jobs need maintainer approval and must be green before merge.

AlexBSoD:
Both points fixed, rebased onto `b42a357`.

**Registration refresh.** The cache moved out of the UI into `linux_setup::IbusRegistrationCache`: a scan is reused for two seconds, our own install/uninstall/reload still invalidate it outright, and the page requests a repaint once the window elapses so an idle screen catches up on its own. The scan is injectable, so three tests cover it — reuse inside the window, a system registration appearing and then disappearing behind the app's back, and invalidation.

**i18n.** Both reload failure paths now go through the catalogs (`ibus_reload_failed_status`, `ibus_reload_lost_status`, en + ru).

576 tests pass, `nix flake check`, `nix fmt -- --ci`, both packages build as 0.3.10, engine `--entropy-check` ok.

av-leschinskiy:
As another NixOS user of an Ergohaven board, I'd be really glad to see this land. Right now the two Linux setup steps in the README are exactly the ones that do not survive on NixOS — the udev rules get wiped on the next activation, and the IBus engine installed into `~/.local/share/ibus/component` is never picked up because `ibus-with-plugins` pins `IBUS_COMPONENT_PATH`. So today it is either a hand-rolled overlay or no Entropy at all.

The nixpkgs package exists but is a release behind, and its module currently fails to evaluate (missing `meta.isIbusEngine`), so an in-repo flake that tracks `main` would be the more reliable path. +1 from me, and happy to test the branch on my own config if that helps.

AlexBSoD:
Rebased onto `60871fe` (v0.3.20); the PR is mergeable again. Two conflicts came from upstream work on the same functions — the new `linux_setup` tests now sit next to mine, and the `linux_setup_task` guard stays in `run_linux_universal_symbols_setup` with the registration reset below it.

The vendored eframe and the new emoji crates needed no flake change. 623 tests, `nix flake check`, `nix fmt --ci`, both packages and `--entropy-check` are green; the workflow still needs an "Approve and run".

kissetfall (approved):
The requested follow-up is complete: the hidraw fallback now uses a dedicated group, Home Manager is evaluated against the real module system, and both x86_64 and aarch64 Nix packages are built natively in CI. All Build and Nix checks pass.

PR #121, opened 2026-07-29
Remove Clippy warnings from layout export

IgorArkhipov:
### Problem

The layout image export UI and render paths passed viewport state, layer context, item geometry, styling, and text coordinates as separate arguments.

Clippy therefore reported six `clippy::too_many_arguments` warnings, adding noise to the warning baseline without identifying a behavior defect.

### Fix

- Reuse `AdaptiveSettingsListViewport` for settings-row geometry, visible rows, and tooltip suppression.
- Group layer render state, item geometry, and rounded-rectangle styling into cohesive values shared by the raster and SVG paths.
- Represent positions and sizes with `egui::Pos2` and `egui::Vec2`, preserving rotation origins and layout units.
- Share key and encoder style constants between raster and SVG output while preserving existing serialized values.

### Verification

- `cargo test --locked layout_image_export -- --test-threads=1` - 7 passed
- `cargo test --locked -- --test-threads=1` - 437 passed
- Binary Clippy warnings decreased from 68 to 62
- All-target Clippy warnings decreased from 79 to 73
- Scoped `rustfmt --edition 2021 --check src/ui/layout_image_export.rs`
- `git diff --check`

Related: #17

### DIFF
--- a/src/ui/layout_image_export.rs
+++ b/src/ui/layout_image_export.rs
@@ -14,6 +14,10 @@ const EXPORT_LAYER_HEADER_H: f32 = 44.0;
 const EXPORT_LAYER_HEADER_H_HIDDEN: f32 = 16.0;
 const EXPORT_LAYER_GAP: f32 = 42.0;
 const EXPORT_MAX_SIDE: f32 = 12_000.0;
+const EXPORT_KEY_RADIUS: f32 = 7.0;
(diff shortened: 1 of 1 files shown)


kissetfall:
Закрываю без мержа как большой lint-only rewrite layout export на старой базе. Пользовательского поведения PR не добавляет, поэтому переносить этот churn не буду.

PR #122, opened 2026-07-30
Remove Clippy warnings from Universal Symbols setup

IgorArkhipov:
### Problem

The Universal Symbols setup helpers for the shared non-macOS page were still compiled on macOS even though macOS now uses its dedicated setup page.

Clippy therefore reported dead-code diagnostics for four helpers, adding noise to the warning baseline without identifying a behavior defect.

### Fix

- Compile the shared setup rows, setup actions, and section label only on non-macOS targets.
- Restrict the simple action-row wrapper to Linux while preserving the shared tooltip-state renderer used by the dedicated macOS page.
- Remove the macOS action block that became unreachable behind the non-macOS setup-actions gate.

### Verification

- `cargo test --locked universal_symbols -- --test-threads=1` - 3 passed
- `cargo test --locked -- --test-threads=1` - 437 passed
- Binary Clippy warnings decreased from 68 to 65
- All-target Clippy warnings decreased from 79 to 78
- Scoped `rustfmt --edition 2021 --check src/ui/universal_symbols_setup.rs`
- `git diff --check`

Related: #17

### DIFF
--- a/src/ui/universal_symbols_setup.rs
+++ b/src/ui/universal_symbols_setup.rs
@@ -380,6 +380,7 @@ impl EntropyApp {
         }
     }
 
+    #[cfg(not(target_os = "macos"))]
     fn draw_universal_symbols_setup_rows(
         &mut self,
         ui: &mut egui::Ui,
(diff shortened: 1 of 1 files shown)


kissetfall:
Закрываю без мержа как устаревший: PR исправляет прежний platform-specific software setup Universal Symbols. Текущая реализация перенесена в RMK и показывается по firmware capability, поэтому этот кодовый путь больше не является актуальным.

PR #123, opened 2026-07-30
Simplify layout geometry helpers

IgorArkhipov:
### Problem

Layout chrome rendering passed five same-typed geometry values as separate arguments, producing a `clippy::too_many_arguments` warning and leaving room for accidental value swaps. An unused keycap rectangle helper also produced both dead-code and argument-count warnings.

### Fix

- Grouped the chrome geometry values in `LayoutChromeGeometry` and destructured them inside the renderer without changing any values or control flow.
- Removed the unused `layout_keycap_rect` helper after confirming it had no callers.

### Verification

- `rustfmt --edition 2021 src/ui/layout_chrome.rs src/ui/layout_shared.rs src/ui/layout_view.rs`
- `cargo check --all-targets --all-features`
- `cargo clippy --all-targets --all-features --message-format short`: binary warnings `68 -> 65`; test warnings `79 -> 76`
- `cargo test --all-features -- --test-threads=1`: 437 passed
- The default parallel test run exposed two existing Vial background-task timing flakes; both passed independently and in the full single-threaded suite.

### DIFF
--- a/src/ui/layout_chrome.rs
+++ b/src/ui/layout_chrome.rs
@@ -6,12 +6,15 @@ impl EntropyApp {
         ui: &mut egui::Ui,
         layout: &KeyboardLayout,
         ctx: &egui::Context,
-        top_base_y: f32,
-        main_tabs_h: f32,
-        layer_bar_h: f32,
-        top_reserved_h: f32,
-        viewport: egui::Rect,
+        geometry: LayoutChromeGeometry,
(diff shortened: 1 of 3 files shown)


kissetfall:
Закрываю без мержа как layout-helper refactor без нового поведения. На актуальной базе перенос этого lint-only churn не оправдан.

PR #126, opened 2026-07-30
chore: simplify app state Clippy lints

IgorArkhipov:
### Problem

Two app-state enums duplicate Default implementations that derives express directly, and a helper repeats an elidable lifetime.

### Fix

Derive the enum defaults and elide the redundant lifetime. No runtime behavior changes.

### Verification

- cargo clippy --locked --all-targets completed successfully.
- Scoped rustfmt --check src/app_state.rs and git diff --check passed.
- No deployment or runtime monitoring is required for this source-only cleanup.

### DIFF
--- a/src/app_state.rs
+++ b/src/app_state.rs
@@ -96,17 +96,12 @@ pub(crate) enum CloseToTrayBehavior {
     Tray,
 }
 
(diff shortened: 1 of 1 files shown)


kissetfall:
Закрываю без мержа как lint-only cleanup старого app state. Пользовательского поведения PR не добавляет.

PR #127, opened 2026-07-30
chore: simplify i18n key variants

IgorArkhipov:
### Problem

Two i18n Key variants repeat the enum name, producing Clippy enum_variant_names warnings.

### Fix

Rename the two internal variants and their UI call sites. Translation identifiers and displayed text remain unchanged.

### Verification

- cargo clippy --locked --all-targets completed successfully; the targeted i18n enum_variant_names diagnostics are zero.
- Scoped rustfmt --check and git diff --check passed.


kissetfall:
Закрываю без мержа как lint-only i18n cleanup на старой базе. Пользовательского поведения PR не добавляет.

PR #128, opened 2026-07-30
chore: remove unused app storage helpers

IgorArkhipov:
### Problem

Two unused app-storage display-name helpers produce dead_code warnings in each compiled target.
### Fix

Remove the unreachable combo and key-override display-name helpers. No persisted data, UI text, or runtime behavior changes.

### Verification

- cargo clippy --locked --all-targets completed successfully; the targeted app_storage dead_code diagnostics are zero.
- Scoped rustfmt --check src/app_storage.rs and git diff --check passed.

### DIFF
--- a/src/app_storage.rs
+++ b/src/app_storage.rs
@@ -513,13 +513,6 @@ pub(super) fn save_combo_colors(colors: &[u32], device_name: &str) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


kissetfall:
Закрываю без мержа как небольшой lint-only cleanup старого app storage кода. Пользовательского поведения здесь нет, перенос не оправдан.

PR #130, opened 2026-07-31
chore: clear UI Clippy warnings

IgorArkhipov:
### Problem

This PR initially added four new `#[allow(clippy::too_many_arguments)]` attributes for established UI helper APIs. Those attributes hid diagnostics rather than resolving them, while a sound parameter-object refactor would affect many existing call sites.

### Fix

- Keep the seven substantive UI and test Clippy fixes in this focused slice.
- Remove the four new allowances; defer the UI API refactor to a separate, reviewable PR.

### Verification / Проверка

- `cargo clippy --locked --all-targets` completes successfully, reporting 62 existing non-blocking warnings, including the four deferred UI-signature diagnostics.
- `cargo test --locked -- --test-threads=1` passed (456 tests) before this documentation-only scope correction; removing lint attributes does not change runtime behavior.
- Scoped `rustfmt --check src/ui_style.rs` and `git diff --check` pass.

### DIFF
--- a/src/app_icon.rs
+++ b/src/app_icon.rs
@@ -55,14 +55,12 @@ fn draw_capsule(
     pixel: &mut [f32; 4],
     x: f32,
     y: f32,
-    ax: f32,
-    ay: f32,
-    bx: f32,
-    by: f32,
+    start: (f32, f32),
+    end: (f32, f32),
     radius: f32,
     color: [f32; 4],
 ) {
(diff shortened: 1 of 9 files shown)


kissetfall:
Закрываю без мержа как UI/Clippy cleanup старого picker-кода. Актуальная ветка уже изменила key picker и capability-gating, поэтому этот diff больше не стоит переносить.

PR #131, opened 2026-07-31
Add GUI chords to Mod-Tap

St1ggy:
## Summary
- add Ctrl+GUI and Shift+GUI Mod-Tap choices
- preserve existing tap-hold behavior for the new modifier chords

## Testing
- cargo test mod_tap_choices_include_gui_chords

### DIFF
--- a/src/keycode_picker_tabs.rs
+++ b/src/keycode_picker_tabs.rs
@@ -411,10 +411,22 @@ impl KeycodePicker {
                 "Ctrl+Alt".into(),
             ),
             (
+                picker_mod_tap_label(0x2900),
+                0x2900,
+                None,
+                format!("Ctrl+{lgui}"),
+            ),
+            (
                 picker_mod_tap_label(0x2600),
                 0x2600,
                 None,
                 "Shift+Alt (LSA)".into(),
+            ),
+            (
+                picker_mod_tap_label(0x2A00),
+                0x2A00,
+                None,
+                format!("Shift+{lgui}"),
             ),
             (
                 picker_mod_tap_label(0x2700),
@@ -506,6 +518,16 @@ mod tests {
         assert_eq!(shift_gui.right_value, Some(0x52BA));
         assert_eq!(shift_gui.label, "OSM\nS+GUI");
         assert_eq!(shift_gui.mod_name, "Shift+GUI");
+    }
+
+    #[test]
+    fn mod_tap_choices_include_gui_chords() {
+        for (base, modifier) in [(0x2900, "Ctrl"), (0x2A00, "Shift")] {
+            assert_eq!(
+                picker_mod_tap_label(base),
+                format!("Hold {modifier}+{}/key", crate::keycode::gui_sym())
+            );
+        }
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


kissetfall (changes requested):
Thanks for adding the missing QMK Mod-Tap chord choices. Please address the UI integration before this is merged:

1. Decode modifier bits 0x09 and 0x0A in the assigned-key tooltip, preferably by reusing the shared modifier decoder. These keycodes currently display "hold for Left modifier" instead of the OS-specific Ctrl+Super/Win/Cmd or Shift+Super/Win/Cmd chord.

2. Preserve and verify the normal Mod-Tap retarget flow. Right-clicking an assigned 0x29xx or 0x2Axx key must reopen the tap-key picker, preserve the modifier chord, and replace only the final tap key. The generic range path appears to support this already, so please add regression coverage for both new chords.

3. Fix the bottom layout hint for these keys. Because these multi-modifier chords do not support handed-side swapping, the current layout-hint branch shows the generic modifier-key action. Any Mod-Tap should show the Mod-Tap tap-key action on right click, independently of whether Ctrl+right-click side swapping is available.

4. Make the new choice-list test verify the actual list contents. The current test only formats two hard-coded bases, so it still passes if the GUI choices are removed. Please expose the choices through a testable helper and assert that 0x2900 and 0x2A00 are present, alongside coverage for the tooltip and right-click behavior above.

Please keep the existing OS-specific Cmd, Win, and Super naming.

src/keycode_picker_tabs.rs
```diff
@@ -486,6 +498,16 @@ mod tests {
         assert_eq!(shift_gui.mod_name, "Shift+GUI");
     }
 
+    #[test]
+    fn mod_tap_choices_include_gui_chords() {
```
IgorArkhipov, line 528: Could we make this test exercise the actual Mod-Tap choices collection too?

That would verify both GUI-chord entries, their keycodes, and held-key labels rather than only the label helper.

kissetfall (changes requested):
The original UI issues are now covered: the branch adds the two choices, preserves tap-key retargeting, fixes the bottom hint, and tests the actual picker contents. The remaining blocker is integration with current `main`.

Please rebase and resolve the conflict by keeping the current modifier ownership instead of restoring a second partial decoder:

- keep `mod_tap_choices` with `0x2900` / `0x2A00` and the new picker, tooltip, retarget, and hint regression tests;
- for full modifier names in tooltips, reuse or extract the shared decoder already owned in `keycode.rs` (the OSM/full-name and modifier-bit paths now cover these GUI chords) rather than extending another local `mod_name` match that can drift;
- preserve the current `main` behavior for the other Mod+Key, One-Shot, RMK-native, and OS-specific Cmd/Win/Super labels.

Once the rebased branch is conflict-free and the current CI matrix is green, the substance of the previous review will be resolved.

kissetfall (approved):
The requested follow-up is complete: GUI-chord Mod-Tap now keeps the correct tap-key hint when side swapping is unavailable, the new choices use shared modifier decoding and OS-specific labels, and regression coverage exercises both 0x2904 and 0x2A04. Build and Nix checks pass.

PR #132, opened 2026-07-31
refactor: group modern text field parameters

IgorArkhipov:
### Problem

Modern text-field helpers accepted long positional argument lists for identity, text, geometry, hints, limits, and alignment. That obscured their shared configuration and produced three `clippy::too_many_arguments` diagnostics.

### Fix

- Introduce `ModernTextField` as the typed parameter object for all text-field rendering inputs.
- Migrate 17 callers without changing field dimensions, limits, alignment, or interactivity.

### Verification

- `cargo clippy --locked --all-targets` completes successfully; diagnostics decrease from 69 to 66 and no text-field argument-count warnings remain.
- `cargo test --locked -- --test-threads=1` passed (456 tests).
- Scoped `rustfmt --check` and `git diff --check` passed.

### DIFF
--- a/src/keycode_picker_macro.rs
+++ b/src/keycode_picker_macro.rs
@@ -306,13 +306,15 @@ impl KeycodePicker {
             if let Some(name) = self.macro_names.get_mut(n) {
                 let resp = crate::ui_style::modern_text_field_sized(
                     ui,
-                    ui.make_persistent_id(("macro_name", grid_id, n)),
-                    name,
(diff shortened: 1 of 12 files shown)


kissetfall:
Закрываю без мержа как широкий UI-refactor без нового пользовательского поведения. Он основан на старой версии общих text-field компонентов и требует отдельного пересмотра уже поверх актуального design-language.

PR #133, opened 2026-07-31
refactor: group settings list row layout

IgorArkhipov:
### Problem

`settings_list_row_with_tooltip` accepted eight positional inputs, including three coupled layout dimensions, at 89 UI call sites. That obscured the row geometry and emitted a `clippy::too_many_arguments` diagnostic.

### Fix

- Add `SettingsListRowLayout` to model a row's content width, height, and control width as one typed value.
- Migrate all 89 callers without changing their label, tooltip, control, or geometry values.

### Verification

- `cargo clippy --locked --all-targets` passes; the current workspace reports 68 non-blocking warnings and none for `settings_list_row_with_tooltip`.
- `cargo test --locked -- --test-threads=1` passed (456 tests).
- Argument equivalence against `origin/main` passed for all 89 callers; scoped `rustfmt --check` and `git diff --check` also passed.

### DIFF
--- a/src/ui/about_device.rs
+++ b/src/ui/about_device.rs
@@ -536,12 +536,14 @@ fn draw_about_rows(
         for row in &rows[list.first_visible_row..list.last_visible_row] {
             crate::ui_style::settings_list_row_with_tooltip(
                 ui,
-                list.row_content_width,
-                list.row_height,
+                crate::ui_style::SettingsListRowLayout::new(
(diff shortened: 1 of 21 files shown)


kissetfall:
Закрываю без мержа: это широкий UI-refactor на базе прежних settings-компонентов и старого Universal Symbols setup. После текущей переделки owner-кода перенос abstraction churn не оправдан.

PR #134, opened 2026-08-01
chore: reduce app and settings Clippy warnings

IgorArkhipov:
### Problem

Ten remaining Clippy diagnostics were spread across app state, update state, platform-specific smart input, deferred loading, firmware settings writes, and Vial tests. Two settings-write diagnostics came from long positional argument lists, where suppressing the lint would hide the warning without improving the API.

### Fix

- Group module-setting writes and readback identity into typed records, preserving every field mapping while removing both `too_many_arguments` diagnostics.
- Apply the remaining behavior-preserving Clippy rewrites for defaults, state initialization, divisibility, platform-gated dead code, eager fallback construction, and test assertions.
- No new `#[allow(clippy::...)]` attributes were added.

### Verification

- `cargo clippy --locked --all-targets` passes; warnings decreased from 60 to 54 for the binary and from 68 to 58 for the test target. All ten selected diagnostics are gone.
- 17 focused tests covering layout-image persistence, typing-trainer punctuation, settings readback and writes, deferred export targeting, background loading, and Vial battery transport passed.
- Scoped `rustfmt --check` and `git diff --check` pass.

### DIFF
--- a/src/app_state.rs
+++ b/src/app_state.rs
@@ -3101,7 +3101,11 @@ fn typing_trainer_number_token(seed: usize, idx: usize) -> String {
 }
 
 fn typing_trainer_should_append_punctuation(seed: usize, idx: usize, word_count: usize) -> bool {
-    idx + 1 < word_count && seed.wrapping_add(idx * 17).wrapping_add(idx / 4) % 5 == 0
+    idx + 1 < word_count
+        && seed
+            .wrapping_add(idx * 17)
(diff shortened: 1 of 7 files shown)


kissetfall:
Закрываю без мержа как cleanup прежней реализации app/settings. В актуальной ветке существенно переделан Smart Input и связанный owner-код, поэтому этот большой lint-only diff уже устарел.

PR #145, opened 2026-08-06
fix: keep sticky layout indicator visible when pinned

IgorArkhipov:
### Problem

When pinned, the sticky Layout Indicator can disappear behind another application's window after Entropy loses focus. It also remains resizable, so its saved size can change while the user expects it to stay fixed.

### Fix

- Keep the pinned viewport at the always-on-top level and disable resizing both when it opens and when pinning changes at runtime.
- On macOS, keep the indicator visible when Entropy deactivates and bring it forward when pinning is enabled.
- Hide the custom resize grip while pinned and restore normal resizing after unpinning.
- Cover the initial and transition window states with pinned and unpinned contract tests.

### Verification

- `cargo test layout_indicator_window` — 2 passed.
- Scoped `rustfmt --check src/ui/layout_indicator_window.rs` and `git diff --check` passed.
- The full suite on the current `origin/main` base remains blocked by existing HID/task timing failures: 486 passed / 18 failed in parallel, and 501 passed / 3 failed single-threaded.


kissetfall (changes requested):
The window-level and resizability changes are structurally sound, and a synthetic merge onto current `main` passed 570/570 tests, i18n, and Clippy. One macOS state transition still needs to be made symmetrical.

Pinning calls `setHidesOnDeactivate: false`, but unpinning sends only `WindowLevel(Normal)` and `Resizable(true)`. The native `hidesOnDeactivate` property therefore remains changed for the lifetime of the viewport: after a pin → unpin cycle, the unpinned indicator does not return to its previous deactivation behavior.

Please restore the native property appropriate to the unpinned state (preferably through one helper that applies both sides of the pin transition, rather than a pin-only native call) and cover the transition contract where possible.

After rebasing onto current `main`, please run native smoke tests on Windows, macOS, and Linux for: initial pinned launch, focus loss, pin → unpin, resize restoration, close/reopen, and persisted state. On macOS, explicitly verify both deactivation behavior and window ordering after unpin.

kissetfall (approved):
The requested pin/unpin state correction is complete on current main. The deferred viewport architecture is preserved, level/resizability transitions are symmetric, macOS ordering is edge-triggered, and the full Build and Nix matrices pass.

PR #160, opened 2026-09-05
fix(matrix-tester): decode switch matrix rows MSB byte first for all firmware

### LINKED ISSUE #156: Matrix Tester highlights the wrong keys on boards with more than 8 columns (matrix row byte order)
### Summary

Matrix Tester decodes the VIA switch-matrix payload with the byte chunks of each row
in the wrong order unless the device name or manufacturer contains "rmk". On any
keyboard with more than 8 matrix columns a pressed switch lights the cap in a
different byte-sized column block, and where that mirrored position falls outside the
declared matrix width the column is never reported at all, so the tester can never
reach 100%. On the 13- and 14-column boards below the shift is exactly 8 columns.

I reproduced this on the same board with Vial-QMK and Vial-RMK firmware, so it is a
host-side decode problem rather than a firmware difference. Boards with 8 columns or
fewer are unaffected because each row then fits in a single byte, which is probably why
this has gone unnoticed.

### Environment / versions

- Entropy v0.3.20; the code paths quoted below are unchanged on current `main` (f60b3df)
- OS: Windows
- Device: OLSK60 v2, USB. Matrix 6 rows x 14 columns with Vial-QMK, 6 rows x 13
  columns with Vial-RMK (the RMK build reads one column through a direct input instead).
  Neither firmware puts "rmk" in its USB strings (manufacturer `techmechkeys`,
  product `OLSK60`).

### Steps to reproduce

1. Connect a Vial keyboard with more than 8 matrix columns (13 or 14 here).
2. Unlock it if required and open Settings -> Matrix Tester.
3. Press a switch in the left-most column (matrix column 0), then one in the `K` column
   (matrix column 8), then `F`, `G`, `H`, `J` (columns 4-7 of the home row).

### Expected behavior

The cap of the switch that was pressed is highlighted.

### Actual behavior / Impact

With the 13-column Vial-RMK build:

| pressed switch | matrix column | cap highlighted in Matrix Tester | column shown |
|---|---|---|---|
| left-most column (Esc / Tab / LCtrl / LShift) | 0 | the `K` column | 8 |
| `K` column | 8 | the left-most column | 0 |
| `G`, `H`, `J` | 5, 6, 7 | nothing at all | - |
| `F` | 4 | nothing (this row has no cap at column 12) | (12) |

The general mapping is: a switch in column 0-4 highlights the cap in column 8-12, a
switch in column 8-12 highlights the cap in column 0-4, and switches in columns 5-7 are
never reported. `F` is the special case in the table: it is in column 4, so it shows up
at column 12, but the OLSK60 home row has no key at column 12, so nothing appears. In
rows that do have a column-12 key the shift is visible directly - for example the `\`
key (row 1, column 12) lights up when the key in column 4 of that row is pressed.

After pressing every switch on the board, columns 5-7 stay dark on every row and the
spacebar can never be marked as tested, so `Tested: n/N` never completes.

With the 14-column Vial-QMK build the same 8-column shift occurs (columns 0-5 <-> 8-13,
columns 6-7 never reported; column 5 lands on column 13, which exists only on the
encoder row of this board), so the home row shows the same symptoms.

### Suggested fix

`get_switch_matrix_with_rmk_byte_order()` (`src/hid_settings.rs`) decodes the payload
with each row's byte chunks in ascending order by default and only reverses them when
`matrix_tester_rmk_byte_order` is set. That flag comes from
`current_device_is_likely_rmk()` (`src/entlayout.rs`), which only checks whether the
device name or manufacturer string contains `rmk`.

The VIA switch-matrix payload, however, always carries the most significant byte of each
row first, for both firmware families:

- QMK `quantum/via.c`, `id_switch_matrix_state`: writes the byte chunks required for
  `MATRIX_COLS` from the highest chunk down, ending with the low byte of
  `matrix_get_row(row)`. On a 13-column board that is two bytes, not four.
- Vial GUI `src/main/python/editor/matrix_test.py`: reads
  `col_byte = len(row_data) - 1 - floor(col / 8)` - unconditionally reversed.
- RMK: emits the same wire order. In 0.8.2 the reversal happens inside
  `MatrixState::read_all()` (`rmk/src/matrix.rs`); on current `main` it has moved to the
  VIA handler (`rmk/src/host/via/mod.rs`, `ViaKeyboardInfo::SwitchMatrixState`), which
  reverses each row's bytes with the comment "Vial wants each row's bytes big-endian
  (QMK matrix_row_t order)". Either way the payload on the wire matches the QMK layout.

So `parse_rmk_switch_matrix_payload()` is the correct decode for QMK and RMK alike, and
the non-reversed `parse_switch_matrix_payload()` is wrong for any Vial device with more
than 8 columns; the name-based heuristic just happens to pick the right branch for
devices that carry "rmk" in their strings. With two bytes per row, the first byte holds
columns 8-15 and the second holds columns 0-7; reading them in ascending order produces
exactly the table above, and bits 5-7 of the second byte (columns 5-7) are never read.

Proposed change: always use the reversed (VIA standard) order in
`parse_switch_matrix_payload_inner()` and drop `matrix_tester_rmk_byte_order` together
with the name-based branch. This is a no-op for boards with 8 columns or fewer and fixes
both Vial-QMK and Vial-RMK boards with wider matrices. I am happy to open a PR if that
is welcome.

### Context

I build and maintain the OLSK60 v2 (RP2040, 60%, 13/14-column matrix) and use Entropy
with both its Vial-QMK and Vial-RMK firmware, which is how the same board ended up being
tested under both.

techmech-keeb:
Fixes #156.

## Summary

`id_switch_matrix_state` was decoded with a byte order chosen from the device
name: `parse_rmk_switch_matrix_payload()` (row bytes reversed) when the device
name or manufacturer contained "rmk", `parse_switch_matrix_payload()` (row bytes
in ascending order) otherwise.

Only the first order is correct. QMK packs each matrix row into
`ceil(cols / 8)` bytes **most significant byte first** — `via.c` writes
`matrix_get_row(row) >> (i * 8)` from the highest chunk down — Vial GUI decodes
`len(row_data) - 1 - floor(col / 8)`, and RMK's VIA response path produces the
same wire order. It is a property of the protocol, not of the firmware family,
so the name check made the default path wrong for every protocol-compliant
board with more than 8 columns that took it.

## Impact

Boards whose matrix fits in 8 columns pack one byte per row, so both paths
agreed and nothing changes for them.

Boards with more than 8 columns had their row chunks read in reverse unless the
device name or manufacturer happened to contain "rmk", so a switch lit the cap
in the mirrored column block, and a column whose mirrored position fell outside
the declared matrix width was not reported at all. On the 13-column build of
the board in #156, columns 0-4 and 8-12 swapped places while columns 5-7
mirrored to the non-existent columns 13-15 and stayed dark.

I cannot tell from this repository which shipped boards are affected. The
dimensions passed to the matrix poll come from the connected device's own Vial
descriptor at connect time (`KeyboardLayout::from_vial_json` reads
`matrix.rows` / `matrix.cols`), and the JSON files under `src/layouts/` are
test fixtures rather than a runtime source — `mod layouts` is `#[cfg(test)]`.
For what it is worth, those fixtures address columns up to 14 for Imperial44
and up to 13 for Omega Point 36, which suggests both have matrices wider than
8 columns, but you are better placed to confirm that against the real
descriptors and USB strings.

The same decode feeds the sticky Layout Indicator window
(`layout_indicator_window.rs` calls `poll_switch_matrix_state`), so its live
key highlighting was affected in the same way and is fixed too.

## Changes

- `parse_switch_matrix_payload()` always reads each row's bytes
  most-significant first. The `reverse_row_bytes` parameter, the
  `parse_rmk_switch_matrix_payload()` wrapper and the
  `get_switch_matrix_with_rmk_byte_order()` entry point are removed.
- The `matrix_tester_rmk_byte_order` flag and the plumbing that carried it
  (`app_state`, `app_init`, `device_connect_apply`, `device_connection`,
  `vial_hid_task`) are removed with it.
- Tests: the existing case is renamed to
  `parses_switch_matrix_with_msb_row_byte_first`, and
  `parses_thirteen_column_switch_matrix_without_column_shift` is added to cover
  a two-byte row where the old default shifted columns.

## Verification

- `cargo test --all-targets`: 694 passed locally on this branch.
- `cargo build --release`: succeeds.
- CI: the repository's Build workflow passes on all four targets on my fork,
  including the Linux `cargo test --all-targets` job —
  https://github.com/techmech-keeb/entropy/actions/runs/33931435969
- Hardware: OLSK60 v2 (6x13) with both Vial-QMK and Vial-RMK, using the Windows
  build from that run. Every switch highlights its own cap (left-most column,
  `K`, `F`/`G`/`H`/`J`), columns 5-7 are detected again, and `Tested: n/N`
  reaches 100%.
- Not verified on hardware: any board other than the OLSK60 v2. The change is
  covered by the added 13-column unit test and by the QMK / Vial GUI wire order
  cited above.

### DIFF
--- a/src/app_init.rs
+++ b/src/app_init.rs
@@ -185,7 +185,6 @@ impl EntropyApp {
             key_override_pick_target: None,
             matrix_tester_pressed: Vec::new(),
             matrix_tester_ever_pressed: Vec::new(),
-            matrix_tester_rmk_byte_order: false,
             sticky_layout_prev_pressed: Vec::new(),
             sticky_layout_pressed_key_layers: Vec::new(),
             sticky_layout_toggled_layers: Vec::new(),
--- a/src/app_state.rs
+++ b/src/app_state.rs
@@ -4687,7 +4687,6 @@ pub struct EntropyApp {
     pub(crate) key_override_pick_target: Option<KeyOverridePickField>,
     pub(crate) matrix_tester_pressed: Vec<bool>,
     pub(crate) matrix_tester_ever_pressed: Vec<bool>,
-    pub(crate) matrix_tester_rmk_byte_order: bool,
     pub(crate) sticky_layout_prev_pressed: Vec<bool>,
     pub(crate) sticky_layout_pressed_key_layers: Vec<Option<usize>>,
     pub(crate) sticky_layout_toggled_layers: Vec<bool>,
--- a/src/hid_parse.rs
+++ b/src/hid_parse.rs
@@ -100,24 +100,22 @@ pub(crate) fn parse_tap_dance_response(resp: &[u8]) -> Result<(u16, u16, u16, u1
     ))
 }
 
-fn parse_switch_matrix_payload_inner(
-    data: &[u8],
-    rows: usize,
-    cols: usize,
-    reverse_row_bytes: bool,
-) -> Vec<bool> {
+/// Decode the VIA `id_switch_matrix_state` payload into a row-major pressed map.
+///
+/// Each matrix row is packed into `ceil(cols / 8)` bytes with the most significant
+/// byte first: the first byte of a row carries its highest columns and the last
+/// byte carries columns 0-7. QMK (`quantum/via.c`), Vial GUI and RMK all emit
+/// and expect this wire order, so it does not depend on the firmware family.
+/// Within the chunk `k = col / 8`, bit `n` is column `8 * k + n`; the chunks
+/// themselves are serialised in descending order.
+pub(crate) fn parse_switch_matrix_payload(data: &[u8], rows: usize, cols: usize) -> Vec<bool> {
     let total = rows * cols;
     let bytes_per_row = cols.div_ceil(8);
     let mut pressed = vec![false; total];
 
     for row in 0..rows {
         for col in 0..cols {
-            let col_byte = col / 8;
-            let row_byte = if reverse_row_bytes {
-                bytes_per_row.saturating_sub(1).saturating_sub(col_byte)
-            } else {
-                col_byte
-            };
+            let row_byte = bytes_per_row - 1 - col / 8;
             let byte_idx = row * bytes_per_row + row_byte;
             let bit_idx = col % 8;
             if byte_idx < data.len() {
@@ -127,15 +125,6 @@ fn parse_switch_matrix_payload_inner(
     }
 
     pressed
-}
-
-pub(crate) fn parse_switch_matrix_payload(data: &[u8], rows: usize, cols: usize) -> Vec<bool> {
-    parse_switch_matrix_payload_inner(data, rows, cols, false)
-}
-
-pub(crate) fn parse_rmk_switch_matrix_payload(data: &[u8], rows: usize, cols: usize) -> Vec<bool> {
-    // RMK MatrixState::read_all emits each row's byte chunks in reverse order.
-    parse_switch_matrix_payload_inner(data, rows, cols, true)
 }
 
 pub(crate) fn parse_vialrgb_supported_effects_payload(
@@ -273,8 +262,8 @@ mod tests {
     }
 
     #[test]
-    fn parses_rmk_switch_matrix_reversed_row_bytes() {
-        let pressed = parse_rmk_switch_matrix_payload(
+    fn parses_switch_matrix_with_msb_row_byte_first() {
+        let pressed = parse_switch_matrix_payload(
             &[0b0000_0010, 0b0000_0100, 0b0000_0001, 0b0000_0000],
             2,
             12,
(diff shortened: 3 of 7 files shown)


kissetfall (approved):
Verified at `58f305693a8722f02ded1c118302448fa2cf0d9a`.

The MSB-byte-first row decoding matches QMK's `matrix_get_row()` payload construction, Vial GUI's existing decoder, and RMK's switch-matrix response. The implementation keeps the change in the protocol decoder and covers multi-byte rows, bit placement, partial final bytes, and invalid payload lengths.

Local full suite: 694 passed, 0 failed. Exact-head fork CI: 4/4 successful.

PR #161, opened 2026-09-05
fix(matrix-tester): keep the board below the header and honour layout options

### LINKED ISSUE #159: Matrix Tester: keyboard is drawn over the header and the "Tested" button, and layout-option variants are drawn on top of each other
### Summary

Two rendering problems on the Settings -> Matrix Tester page (screenshot attached):

1. The keyboard is drawn on top of the page header: the description line is crossed by
   the top key row and the `Tested: n/N` button is covered by the board, so the
   progress count cannot be read. The button still reacts to clicks and resets
   progress; only its appearance is hidden underneath the keycaps.
2. Keys that belong to different layout options are all drawn at once, so alternative
   spacebar variants land on top of each other. The bottom row shows the 6.25U
   spacebar variant painted over the split-space variants, which are hidden underneath.

Both are independent of #156, which is about decoding the matrix payload; these two are
purely about how the board is drawn. I have implemented both on my fork and verified
them on hardware, so this is a report plus an offer rather than only a bug report.

### Environment / versions

- Entropy v0.3.20; the code paths quoted below are unchanged on current `main` (c8b5297)
- OS: Windows, default window size (1200x700)
- Device: OLSK60 v2, 15u x 6.25u layout, layout options `Spacebar` (4 values) and
  `RotaryEncoder` (2 values) in `vial.json`

### Steps to reproduce

1. Connect a Vial keyboard whose layout is wider than it is tall and whose `vial.json`
   defines layout options (OLSK60 v2 here).
2. Open Settings -> Matrix Tester in the default 1200x700 window.
3. Look at the description line and the `Tested: n/N` button, then press the space key.

### Expected behavior

- The title, the description and the `Tested: n/N` button stay readable above the
  board, and the board stays inside the area below them.
- Only the caps of the currently selected layout option are drawn, the same way the
  Layout page draws them.

### Actual behavior / Impact

1. The description line is crossed by the top row of caps and the `Tested` label is
   readable only where a gap between caps happens to fall; the count is behind a cap.
2. All spacebar variants are painted together and the last one defined (6.25U) wins
   visually, so caps that exist on the physical board are not visible, pressing the
   space key lights a wide cap spanning keys that were not pressed, and the
   `Tested: n/N` denominator counts every variant separately for the same matrix
   position.

### Suggested fix

1. Board overlap. `draw_matrix_tester_settings()` (`src/ui/matrix_tester.rs`) places
   the title, the description and the `Tested` button relative to `content_rect`,
   which already starts at panel top + `LAYOUT_TOP_RESERVED_H` (104 px,
   `src/ui/settings_shell.rs` / `src/ui/layout_chrome.rs`). The key geometry, however,
   is built from the whole panel rect:

   ```rust
   let viewport = egui::Rect::from_min_max(
       ui.min_rect().min,
       egui::pos2(ui.min_rect().left() + ui.available_size().x, ui.max_rect().bottom()),
   );
   let geometry = layout_geometry(ui.ctx(), layout, viewport, clamp_ui_scale(...));
   ```

   `layout_geometry()` centres the board between `viewport.top() + LAYOUT_TOP_RESERVED_H`
   and `viewport.bottom() - LAYOUT_BOTTOM_RESERVED_H`, so the 104 px are reserved from
   the panel top again instead of from below the tester header. `board_rect`, computed
   a few lines above at `content_rect.top() + 104.0`, is only used to centre the
   "not supported" / "connect a keyboard" messages. With the default window and a
   15u x 6.25u board the scale is capped by `responsive_layout_max_scale()`, the board
   is about 388 px tall and its top edge lands near panel top + 170 px, while the
   `Tested` button occupies panel top + 177..207 px. On higher display scaling the
   board grows and covers the description line as well. Proposed change: build the
   geometry from `board_rect` (for example via `layout_geometry_with_reserved()` with no
   extra top/bottom reserve) so the board stays inside the area the page already
   reserves for it.

2. Layout-option variants. The drawing loop iterates over all of `layout.keys` without
   the visibility filter that `draw_layout()` applies through
   `Self::layout_key_visible()` (`src/ui/layout_view.rs`), and `total_keys` counts the
   same unfiltered list. Proposed change: apply the same filter for drawing and for the
   tested counter, and pass it to the filtered geometry helper so the board is sized
   from the visible keys only.

Note: in the screenshots the `Tested:` label is readable but the count itself sits
behind a keycap, which is issue 1 of this report.

### Fix and verification

Both are implemented on my fork and verified on an OLSK60 v2 under Vial-QMK and
Vial-RMK, using a Windows build from the repository's own Build workflow:

- the description line and the `Tested: n/N` count are readable, with the board below
  them;
- the bottom row shows only the caps of the selected `Spacebar` option;
- the `Tested` button still resets progress when clicked.

I am happy to open a PR for either or both if that is welcome. I would send them as two
separate commits so each can be taken or dropped on its own.

### Context

I maintain the OLSK60 v2 and hit this while running Matrix Tester on it under Vial-QMK
and Vial-RMK. The screenshots were taken on the RMK build; the layout and the overlap
are the same on QMK.

techmech-keeb:
Fixes #159. Two independent rendering defects in the Matrix Tester, kept as two
commits as requested.

## 1. The board covered the header

`fix(matrix-tester): keep the board inside its own area below the header`

The tester draws its title, description and `Tested` button inside
`content_rect`, but it built the key geometry from the whole panel viewport.
`layout_geometry()` reserved `LAYOUT_TOP_RESERVED_H` from the panel top, which
lands exactly on `content_rect.top()` — nothing accounted for the tester's own
header and status block below that. The board therefore started a full header
height too high, and at the default window size its caps covered the
description line and the `Tested` button, leaving the text unreadable and the
button unclickable.

`board_rect` was already computed for the area between that header and the
bottom hint, but it was only used to centre the placeholder messages. The fix
passes that already-trimmed rectangle to the existing
`layout_geometry_with_reserved_and_filter()` with no additional top or bottom
reserve, so the caps stay between the header and the hint.

The tester was the only caller of `layout_geometry()`, so this commit removes
that wrapper; `layout_geometry_with_reserved()` stays until the second commit
routes the tester through the filtering variant and leaves it unused. The other
on-screen views already call `layout_geometry_with_reserved_and_filter()`
directly.

## 2. Layout-option variants drew on top of each other

`fix(matrix-tester): draw and count only keys of the selected layout options`

The tester drew every key in the layout definition, including the variants of
each layout option. Variants share matrix positions, so they were painted on
top of each other and the last one defined won. On a board with a split/unsplit
spacebar row this meant caps that exist on the physical keyboard were hidden,
pressing one switch lit a cap spanning several keys, and the `Tested`
denominator counted each variant separately so the counter could never reach
100%.

The fix applies `layout_key_visible()` to the counter, the geometry and the
drawing loop. That is the same filter the Layout page, the layout indicator
preview and the image export already apply; the Matrix Tester was the only
view drawing every variant.

Note that `layout_key_visible()` covers more than `layout_options_value`: for a
module-owned encoder press key it defers to the existing module-settings rule
(`module_settings_encoder_visible`), so setting a module to None, Trackball or
Touchpad now hides that module's press key here and drops it from the `Tested`
denominator, exactly as it already does in the other views. Tell me if you would
rather the tester ignore that rule and probe every matrix position regardless of
the module configuration — that would be a one-line change to
`layout_condition_visible()`, but it would make the tester the only view with
its own visibility rule.

Encoders are excluded from the geometry because the tester does not draw
them. Routing the tester through
`layout_geometry_with_reserved_and_filter()` leaves
`layout_geometry_with_reserved()` unused, so this commit removes it.

## Verification

- `cargo test --all-targets`: 693 passed locally on this branch.
- Each commit builds on its own: `cargo check --all-targets` passes at both
  the first and the second commit, so the branch stays bisectable.
- `cargo build --release`: succeeds.
- CI: the repository's Build workflow passes on all four targets on my fork,
  including the Linux `cargo test --all-targets` job -
  https://github.com/techmech-keeb/entropy/actions/runs/33945567691
- Hardware: OLSK60 v2 (6x13). The description and the `Tested` button are
  readable and clickable at the default window size, every physical cap is
  drawn exactly once, and `Tested: n/N` reaches 100%.
- Not verified on hardware: the module-owned encoder press-key path described
  above. The OLSK60 v2 has no interchangeable Encoder / Trackball / Touchpad
  module, so that branch of `layout_key_visible()` is covered only by the
  existing unit test `module_settings_own_the_conditional_encoder_press_key`.

### DIFF
--- a/src/ui/layout_shared.rs
+++ b/src/ui/layout_shared.rs
@@ -20,30 +20,6 @@ pub(crate) fn responsive_layout_max_scale(ctx: &egui::Context, viewport: egui::R
     1.0 + 0.35 * t
 }
 
-pub(crate) fn layout_geometry_with_reserved(
-    ctx: &egui::Context,
-    layout: &KeyboardLayout,
-    viewport: egui::Rect,
-    ui_scale: f32,
-    top_reserved: f32,
-    bottom_reserved: f32,
-    fit_margin: f32,
-    max_scale_override: Option<f32>,
-) -> LayoutGeometry {
-    layout_geometry_with_reserved_and_filter(
-        ctx,
-        layout,
-        viewport,
-        ui_scale,
-        top_reserved,
-        bottom_reserved,
-        fit_margin,
-        max_scale_override,
-        |_| true,
-        |_| true,
-    )
-}
-
 pub(crate) fn layout_geometry_with_reserved_and_filter(
     ctx: &egui::Context,
     layout: &KeyboardLayout,
--- a/src/ui/layout_shared.rs
+++ b/src/ui/layout_shared.rs
@@ -20,24 +20,6 @@ pub(crate) fn responsive_layout_max_scale(ctx: &egui::Context, viewport: egui::R
     1.0 + 0.35 * t
 }
 
-pub(crate) fn layout_geometry(
-    ctx: &egui::Context,
-    layout: &KeyboardLayout,
-    viewport: egui::Rect,
-    ui_scale: f32,
-) -> LayoutGeometry {
-    layout_geometry_with_reserved(
-        ctx,
-        layout,
-        viewport,
-        ui_scale,
-        LAYOUT_TOP_RESERVED_H,
-        LAYOUT_BOTTOM_RESERVED_H,
-        LAYOUT_FIT_MARGIN,
-        None,
-    )
-}
-
 pub(crate) fn layout_geometry_with_reserved(
     ctx: &egui::Context,
     layout: &KeyboardLayout,
--- a/src/ui/matrix_tester.rs
+++ b/src/ui/matrix_tester.rs
@@ -162,18 +162,33 @@ impl EntropyApp {
             self.prompt_if_vial_locked_for_matrix_poll();
         }
 
-        let total_keys = layout.keys.len();
-        let tested_count = layout
-            .keys
-            .iter()
-            .filter(|key| {
-                let idx = key.row as usize * layout.cols + key.col as usize;
-                self.matrix_tester_ever_pressed
-                    .get(idx)
-                    .copied()
-                    .unwrap_or(false)
-            })
-            .count();
+        // Only the caps of the currently selected physical configuration exist
+        // on the board: alternative layout-option variants share matrix
+        // positions and must not be drawn or counted on top of each other, and
+        // layout_key_visible() also hides a module-owned encoder press key when
+        // its module is set to something other than an encoder.
+        let mut total_keys = 0;
+        let mut tested_count = 0;
+        for key in &layout.keys {
+            if !Self::layout_key_visible(
+                &self.module_settings,
+                layout,
+                key,
+                self.layout_options_value,
+            ) {
+                continue;
+            }
+            total_keys += 1;
+            let idx = key.row as usize * layout.cols + key.col as usize;
+            if self
+                .matrix_tester_ever_pressed
+                .get(idx)
+                .copied()
+                .unwrap_or(false)
+            {
+                tested_count += 1;
+            }
+        }
 
         crate::ui_style::allocate_ui_at_rect(
             ui,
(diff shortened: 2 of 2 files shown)


kissetfall (approved):
Verified at `3018d1b0e5f252c4aa15200c41fdab73c041c00e`, including a clean local merge with the new `main` after #160.

The fix follows the existing rendering path: `layout_key_visible` is now applied consistently to progress counting, geometry, and drawing, while the board geometry uses the already-trimmed `board_rect` below the header. The removed wrapper is no longer used elsewhere.

Merged-tree local suite: 694 passed, 0 failed. Exact-head fork CI: 4/4 successful. The OLSK60 v2 hardware result covers the reported overlap and layout-option behavior.

Non-blocking test gap: there is no automated visual/layout-option regression test, and the module-owned encoder-visibility branch was not hardware-tested. Neither changes the merge verdict.

PR #19, opened 2026-06-19
Stabilize macOS HID initialization on the main thread

IgorArkhipov:
### Summary
Stabilize macOS HID initialization by initializing hidapi on the main thread before UI startup.

This targets the Apple Silicon crash observed after moving Entropy.app to Applications, where hidapi's Darwin IOHIDManager can bind to an unsafe first run loop. It also preserves the Rosetta HID scan block.

### Verification
- Local: `git diff --check`
- Local: `cargo build --release`
- Local: `cargo test`
- Local: `scripts/build_macos_app.sh`
- Remote: GitHub Actions `Build` run: https://github.com/ergohaven/entropy/actions/runs/27808718588.

Fixes #8

### Кратко
Стабилизирует macOS HID инициализацию: hidapi инициализируется на main thread до запуска UI.

Изменение закрывает ошибку на Apple Silicon, которая воспроизводился после переноса Entropy.app в папку Applications, когда Darwin IOHIDManager в hidapi мог привязаться к небезопасному first run loop. Rosetta HID scan block сохранен.

### Проверка
- Локально: `git diff --check`
- Локально: `cargo build --release`
- Локально: `cargo test`
- Локально: `scripts/build_macos_app.sh`
- Remote: GitHub Actions `Build` run: https://github.com/ergohaven/entropy/actions/runs/27808718588.

Закрывает #8

PR #162, opened 2026-09-08
feat(modules): label the auto layer deactivate-on-key setting

iakunin:
## What

K:04 firmware `v0.1.10` exposes a new Vial QMK setting, `Auto layer deactivate on key` (qsid 335), which drops the pointing auto layer as soon as a non-mouse key is pressed instead of waiting out the inactivity timeout.

Entropy builds the module settings page straight from the device definition (`module_settings_groups`), so the row appears on its own once firmware advertises it — but with the raw English title and the generic `Firmware setting: {field}` tooltip. This adds the label and tooltip.

## Changes

- `src/ui/module_settings.rs` — catalog entry mapping the title to its label/tooltip keys
- `src/i18n.rs` — `static_catalog_key` entry
- `i18n/en.toml`, `i18n/ru.toml` — label + tooltip strings
- one test feeding a definition JSON through the real `module_settings_groups` pipeline, asserting the field lands in the Auto Layer group as a boolean row, visible in every pointer mode, and localized in both languages

No production code references qsid 335 — resolution is by title string, so this stays firmware-driven.

## Firmware side

Requires ergohaven/rmk#19, which allocates qsid 335 and implements the behaviour. Without it the row simply does not appear; with an older Entropy the row appears untranslated. No hard dependency in either direction.

## Verification

- `cargo test --bin entropy` — 657 passed, 0 failed
- `cargo fmt --check` — no diffs in any touched file (pre-existing drift in `keycode.rs`, `layout_hints.rs`, `layout_indicator_window.rs` left alone)
- `scripts/check_i18n.py` — clean
- Confirmed end-to-end on a K:04 Mini running the matching firmware

https://claude.ai/code/session_017y8Qh1NR26NxtP2wR1nPLu


i18n/en.toml
```diff
@@ -1328,6 +1328,7 @@ auto_layer_sniper_tooltip = "Automatically switch to the selected layer in snipe
 auto_layer_scroll_tooltip = "Automatically switch to the selected layer in scroll mode"
 auto_layer_text_tooltip = "Automatically switch to the selected layer in text mode"
 auto_layer_timeout_tooltip = "How long the firmware keeps the auto layer active after module movement stops"
+auto_layer_deactivate_on_key_tooltip = "Turn the auto layer off as soon as any non-mouse key is pressed"
```
kissetfall, line 1331: **P3, nonblocking:** `any non-mouse key` is broader than the firmware contract: layer-switch actions, ordinary macro triggers, pointer-mode User actions and bare Again include intentional exceptions. Please qualify the wording or mention those exceptions rather than promising every physical non-mouse key. The corresponding Russian tooltip in `i18n/ru.toml` should stay consistent. This is a copy clarification, not a request to change the firmware classifier.

kissetfall (approved):
The Entropy changes follow the existing firmware-driven settings, capability filtering and localization paths. No blocking functional issue found at this head. I verified 37 exact-head module-settings tests, 2 supplemental compatibility/localization tests, 1 old-client fallback test, and the i18n checker. The four changed paths also merge cleanly with current main `004dbff2`.

This client-side localization change can land independently of the firmware behavior fix in ergohaven/rmk#19: older firmware does not expose the field, and older Entropy already handles it through the generic settings UI. Please see the nonblocking tooltip suggestion inline. CI still needs to pass before merge.

